Issue #542, opened 2015-01-26
RestAPI: URL for restricted files should be hidden


SarahBA:
- URL for restricted files should be hidden.
Another little bug to fix: 
- Return an empty list (instead of error) for empty communities.

SarahBA:
fixed by https://github.com/EUDAT-B2SHARE/b2share/pull/551

PR #551, opened 2015-02-02
Restapi: fixing hidden files issue & removing existing deposition folder

SarahBA:


SarahBA:
fixing https://github.com/EUDAT-B2SHARE/b2share/issues/542 & https://github.com/EUDAT-B2SHARE/b2share/issues/552

invenio/b2share/modules/b2deposit/restful.py
```diff
@@ -66,6 +68,7 @@
     'domain':               ('980__a', False), #\ same marctag
     'resource_type':        ('980__a', True),  #/ same marctag
     'publication_date':     ('260__c', False),
+    'publisher':            ('8560_f', False),
```
emanueldima: This is not the publisher field. The publisher field is 260/b, and in our case always the url of b2share. What are we trying to do here?

invenio/b2share/modules/b2deposit/restful.py
```diff
@@ -272,10 +283,12 @@ def post(self, oauth, **kwargs):
         CFG_B2SHARE_UPLOAD_FOLDER = current_app.config.get(
                                 "CFG_B2SHARE_UPLOAD_FOLDER")
         deposit_id = uuid.uuid1().hex
-        upload_dir = os.path.join(CFG_B2SHARE_UPLOAD_FOLDER, deposit_id)
+        deposit_id_folder = 'temp/' + deposit_id
```
emanueldima: I don't understand why we need to put stuff into a ...../temp/ folder.

SarahBA:
By "publisher" I mean the person who published the record. If it previously used as our url then I should better use user_email_address or something similar.

emanueldima:
This pull request is too large for what it does. I suggest to close this one, do a rebase, then resubmit it.

emanueldima:
@SarahBA The UI uses the 'Uploaded by' tag for this field, it is more appropriate in my opinion.

SarahBA:
ok, will send a new pull request

Issue #566, opened 2015-02-16
Check against duplication of domain field names when invenio starts


emanueldima:
Adding new domains means adding new fields, and the more domains we add the more probable a clash becomes.

noumar:
The problem I'm having is with ```generic``` and ```domain``` clashing, not between domains.

If you create a field in a domain that already exists within generic, e.g ```language```, you get a cryptic error: ```TypeError: Do not know how to convert multiple-column properties currently```.
This is completely useless and not helping to find the name of the field.

To my understanding the metadata fields are just key-value pairs stored separately for each record so they should not clash as one record can't belong to several domains.

noumar:
@emanueldima have you had the problem that field names in two domains clash?

emanueldima:
@noumar not yet; can you post the domain/field names that you are trying to add?

noumar:
See #540 where RDA wants Dublin Core fields for their community. In practice this means 11 are already provided by ```Generic``` but 5 need to be ```RDA``` custom, unless we extend ```Generic```. This takes time to match up manually, but the comparing could well be done in ```model.py```. Currently ```model.py``` just adds ```Generic``` + ```Community``` fields together and "duplicate fields" somehow end up on top of each other and that causes the weird error.

noumar:
@emanueldima I have added some checks to ```model.py``` that checks field names between Generic and Domain, and duplicates within Domain. I can push the changes today, but to this issue or a new one?

emanueldima:
This one, but please don't close the issue as I might have some code for it later

noumar:
@emanueldima please have a look at my suggestions when you can.

PR #568, opened 2015-02-19
WIP: #566 duplicate fields

noumar:
#566

### DIFF
--- a/invenio/b2share/modules/b2deposit/b2share_model/model.py
+++ b/invenio/b2share/modules/b2deposit/b2share_model/model.py
@@ -15,6 +15,8 @@
 ## along with B2SHARE; if not, write to the Free Software Foundation, Inc.,
 ## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
+import collections
+
 from invenio.ext.sqlalchemy import db
 from flask import current_app
 from datetime import date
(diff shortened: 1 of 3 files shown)


invenio/b2share/modules/b2deposit/b2share_model/model.py
```diff
@@ -95,13 +96,11 @@ def __init__(self):
...
         self.field_args['publisher'] = {
-            'hidden': True,
-            'value': self.publisher_default
-            # 'description':
-            # 'Here should be stored the site that will host the BE2Share ' +\
-            # 'container, so that in case of access problems, people can ' +\
-            # 'be contacted. This element can be created automatically ' +\
-            # 'dependent on the centre.'
+            'value': self.publisher_default,
+            'description': 'Here should be stored the site that will host the B2SHARE '
```
emanueldima: This description must be changed

invenio/b2share/modules/b2deposit/b2share_model/model.py
```diff
@@ -140,10 +139,8 @@ def __init__(self):
                            'were relevant in the creation of the resource.'
         }
         self.field_args['language'] = {
-            'hidden': True,
             'value': self.language_default,
-            # 'description':
-            # 'The name of the language the document is written in.'
+            'description': 'The name of the language the document is written in.'
```
emanueldima: The Linguistics domain also has a language_code field. We must avoid having 2 different fields for the same thing, especially if they can be filled in by the user with contradictory information.

emanueldima:
I agree with the patch in general, but see the line comments.

invenio/b2share/modules/b2deposit/b2share_model/model.py
noumar: Agree, but what "should" the field represent? The name, description, common use and our use are not consistent. We assign it automatically a "service instance url" but it's unclear to me why so, and e.g. an article could already have been published in a magazine to which the uploader wants it to refer.

@cjhak can you maybe give some input on this? Specifically use and description.

invenio/b2share/modules/b2deposit/b2share_model/model.py
```diff
@@ -95,13 +96,11 @@ def __init__(self):
...
         }
         self.field_args['publisher'] = {
-            'hidden': True,
-            'value': self.publisher_default
-            # 'description':
-            # 'Here should be stored the site that will host the BE2Share ' +\
-            # 'container, so that in case of access problems, people can ' +\
-            # 'be contacted. This element can be created automatically ' +\
-            # 'dependent on the centre.'
+            'value': self.publisher_default,
```
noumar: @emanueldima do you know how to handle the value so it isn't stored as `None` in the DB? It only occurs if the user empties the prefilled field.

invenio/b2share/modules/b2deposit/b2share_model/model.py
noumar: Agree, but this doesn't seem to be used, at least not visible, anywhere on the record page. So what's been the purpose of the field?

emanueldima:
I'll merge this one and try to fix the problems later as it's important to have the rda domain integrated fast

PR #571, opened 2015-02-23
misc fixes related to optional basic fields and unicode file names

emanueldima:
see #566 

adds a better description for the publisher field and other fixes

### DIFF
--- a/invenio/b2share/modules/b2deposit/b2share_model/metadata/linguistics_lang_codes.py
+++ b/invenio/b2share/modules/b2deposit/b2share_model/metadata/linguistics_lang_codes.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 lang_codes = [
+    ('eng', 'English'),
     ('aaa', 'Ghotuo'),
     ('aab', 'Alumu-Tesu'),
     ('aac', 'Ari'),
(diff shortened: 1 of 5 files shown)


invenio/b2share/modules/b2deposit/b2share_model/metadata/linguistics_lang_codes.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 lang_codes = [
+    ('eng', 'English'),
```
noumar: @emanueldima why did you duplicate the `eng` row? You should not modify the closed vocabulary. If you intend to set a default value then this should be done modifying `HTML5ModelConverter.py:180` to support that, as the order or further processing of the list might be added/done. But if not, then at least remove the duplicate `eng` row that appears later. There is a new 2015 revision of ISO 639-3 out, and when rebuilding the pair-list this might easily be overseen.

emanueldima: @noumar You are correct that duplicating a field is not the best option, but the previous behaviour was much worse. I don't know exactly how to modify the model converter, do you?

emanueldima: @noumar ... just created a new issue for this: #573

PR #578, opened 2015-02-27
Add 'discipline' metadata field

### LINKED ISSUE #541: "Discipline" field in the generic schema
The following request is from the B2FIND team:

"Within b2find we are using the facet discipline. Only a closed list of
entries are possible here, as defined in
http://en.wikipedia.org/wiki/Outline_of_academic_disciplines
Is it possible to have such a field (with the same list) within the generic
metadata scheme of b2find? (it should be possible to select 0, 1 or more out of
those in the list).
A subset of those disciplines is also available in git within
https://github.com/B2FIND/md-mapping/blob/master/mapfiles/b2find_disciplines.tab"

noumar:
Closes #541

### DIFF
--- a/invenio/b2share/modules/b2deposit/b2share_model/model.py
+++ b/invenio/b2share/modules/b2deposit/b2share_model/model.py
@@ -16,14 +16,17 @@
 ## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
 import collections
+import csv
+import os.path
 
 from invenio.ext.sqlalchemy import db
 from flask import current_app
 from datetime import date
 
+CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
+
 
 class FieldSet:
-
     def __init__(self, name, basic_fields=[], optional_fields=[]):
         self.name = name
         self.basic_fields = basic_fields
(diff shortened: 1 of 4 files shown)


invenio/b2share/modules/b2deposit/b2share_model/model.py
```diff
@@ -16,14 +16,17 @@
...
 
 import collections
+import csv
+import os.path
 
 from invenio.ext.sqlalchemy import db
 from flask import current_app
 from datetime import date
 
+CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
```
emanueldima: wouldn't be better to use pkg_resources for that? see http://stackoverflow.com/questions/779495/python-access-data-in-package-subdirectory

noumar: Yes, I agree it would but I didn't use it because of the reasons below:

1. http://stackoverflow.com/questions/779495/python-access-data-in-package-subdirectory#comment25649364_5601839
1. `%time import pkg_resources
CPU times: user 42.1 ms, sys: 12.8 ms, total: 54.8 ms
Wall time: 71.2 ms`
1. `%time import os.path
CPU times: user 7 µs, sys: 0 ns, total: 7 µs
Wall time: 10 µs`
1. `__package__` is broken in Python 3.0 - 3.2

Issue #544, opened 2015-01-28
RestAPI body (JSON) structure proposal


dblommesteijn:
Testing the RestAPI especially the response body (JSON) is difficult to test because sometimes the API returns: Empty String, dict (`{}`) or list (`[]`) #489. I'd like to propose a base structure of the API JSON response so it can integrated more easily.

Structuring the responses, as described below, will always respond with a valid JSON (and dict in programming languages). For future applications we can even autobox these structures to form valid objects.

**Deposit**
```json
{ "deposit": ... }
```

**Deposits**
```json
{ "deposits": [
    { "deposit": ... }
] }
```

**Empty payload**
```json
{}
```

dblommesteijn:
It would be nice if we have a JSON payload incorporating the following:

```json
# wrapper around all deposits
{ "deposits": [ depositX, depositY] }
# wrapper for a single deposit
{ "deposit": {} }
```
This way users can automatically box (wrap/ create) objects.

dblommesteijn:
Number of deposits in json proposal:
```json
{ "deposits": [
    { "deposit": ... }
] , 
   "deposits_no": n 
}
```

@emanueldima @SarahBA perhaps we could use some json payload as described above

PR #587, opened 2015-03-16
JSON payload Response

SarahBA:
Fixing https://github.com/EUDAT-B2SHARE/b2share/issues/544


invenio/b2share/modules/b2deposit/restful.py
```diff
@@ -205,9 +206,14 @@ def get(self, oauth, domain_name, **kwargs):
         # get domain id from domain name
         domain = Bib98x.query.filter_by(value=domain_name).first()
         if domain is None:
-            abort(404, status=404,
-                  message="Please try a valid domain name: " +
-                         ", ".join(metadata_classes().keys()))
+            if domain_name not in metadata_classes().keys():
```
emanueldima: This check should be moved up, before line 206 (before asking the database). It is part of the initial checks of the function, right?

invenio/b2share/modules/b2deposit/restful.py
```diff
@@ -216,13 +222,28 @@ def get(self, oauth, domain_name, **kwargs):
...
         stop = start + page_size
+
         for record_id in record_ids[start : stop]:
             record_details = get_record_details(record_id, curr_user_email)
             record_list.append(record_details)
-        if stop < len(record_ids):
-            record_list.append('...') # continuation indicator
-        return record_list
 
+        if len(record_list) == 0:
```
emanueldima: All this long list of ifs seems unnecessary to me. Isn't it equivalent with:

`if len(record_list) == 0: return jsonify({}); else: return jsonify(Deposits=record_list)` ?

but it might be that I'm missing something, please comment

invenio/b2share/modules/b2deposit/restful.py
```diff
@@ -216,13 +222,28 @@ def get(self, oauth, domain_name, **kwargs):
...
+        elif len(record_ids) == 1 and page_offset > 0:
+            resp = jsonify({})
+        elif len(record_ids) > 1 and page_offset == 0:
+            resp = jsonify(Deposits = record_list)
+        elif len(record_ids) > 1 and page_offset > 0 and len(record_list) > 0:
+            resp = jsonify(Deposits = record_list)
+        elif len(record_ids) > 1 and page_offset > 0 and len(record_list) == 0:
+            resp = jsonify({})
+
+        resp.headers['Number of Deposits'] = len(record_ids)
```
emanueldima: Is it customary to return this type of information in a response header? It's a new thing for me.

invenio/b2share/modules/b2deposit/restful.py
SarahBA: I think it’s ok. Previously, if user asked for an empty domain he was getting the message for entering wrong domain name, which is not the correct message. Instead I think he should get an empty list. Now if domain is “None”, it checks for two cases: if it is not in the domain list (metadata_classes().keys()) then he’ll get an error message, else he will get an empty list.

invenio/b2share/modules/b2deposit/restful.py
SarahBA: These are the cases
1- if it is empty => JSONIFY({})
2- if there is only one deposit => there are two cases:
.. 1- if user requests whit no page_offset -> JSONIFY(Deposit)
.. 2- if user added (unnecessary page_offset ) ->  JSONIFY({})
3- if there are more deposits:
.. 1- if user asked for first page_offset -> jsonify(Deposits)
.. 2- if user for other page_offset:
.... * if there is still more deposits in the requested page_offset: ->  jsonify(Deposits)
.... * if no more deposits in the requested page_offset, should return empty ->  jsonify({})

emanueldima: You are correct that these cases exist, but my code above has exactly the same effect and is much shorter. Less code is almost always better. I don't see the point of having 6 ifs when you can return the same result with just one.

dblommesteijn:
I've a more fundamental question, why would one need to know the number of deposits? The JSON payload can easily be queried for such information.

jirikuncar:
> I've a more fundamental question, why would one need to know the number of deposits?

What if you have many deposits? Are you planning to support pagination?

dblommesteijn:
We've already got pagination, what's your question?

jirikuncar:
> why would one need to know the number of deposits?

If you have pagination then you already have "number of deposits" in the response.

PR #606, opened 2015-04-08
Vagrant environment update

emanueldima:
Commented out the yum system update in the provisioning scripts, which seems to fix the vagrant sync issues


dblommesteijn:
I'm not sure about the custom `./install.sh` script, instead of `vagrant up`.

emanueldima:
@dblommesteijn I agree, we should automate this. New issue?

dblommesteijn:
You can just fix it with this PR I think.

emanueldima:
@dblommesteijn Won't happen until the end of next week. This PR fixes a big and different issue

dblommesteijn:
I've got an issue with the installation of current `master` install, will this PR fix it?

PR #614, opened 2015-04-17
Update latest_deposits.py

llehtine:
this should fix the latest deposits showing 4 first deposits


emanueldima:
Yes, I can see why

dblommesteijn:
Why did we merge a pull-request that was failing Travis?

emanueldima:
There's no way this fix could have crashed travis. I think travis didn't like the fact that I hastily removed the branch, let's restart it.

emanueldima:
It's possible that an update of some python package broke the travis build; this upstream fix might be related: https://github.com/inveniosoftware/invenio/commit/f7c3f768eefc658a70452eaa617a2035175c25b7

PR #616, opened 2015-04-20
Revert "Update latest_deposits.py"

emanueldima:
Reverts EUDAT-B2SHARE/b2share#614

I don't really want to revert the fix, I just want to see if travis fails with the old code or not.


emanueldima:
Travis still fails, so it looks like Lassi's commit has nothing to do with it. I will close this PR.

PR #631, opened 2015-05-04
hiding Unity login option, for now

emanueldima:


cjhak:
@emanueldima We should merge this fix to 1.4, not to master, at this point.

emanueldima:
Correct.

PR #665, opened 2015-06-02
Fixed admin page (double inclusion of js scripts)

### LINKED ISSUE #664: The admin interface does not work
On a development VM the admin interface at `0.0.0.0:4443/admin` does not work. The console reports this error: `Uncaught Error: See almond README: incorrect module build, no module name`

emanueldima:
`base/scripts.html` was included twice, once in the `header` block of `admin_base.html` and once in the `_bottom_assets` block of the `page_base.html`

 fixes #664


emanueldima:
@nharraud This fix should probably be also ported to invenio

Issue #667, opened 2015-06-05
Deny uploading empty files via REST API


emanueldima:
This is a temporary measure because https://github.com/EUDAT-B2SHARE/b2share/issues/575 is delayed.

dblommesteijn:
@emanueldima Is this already fixed in a pull-request?

emanueldima:
No, not yet. You can have it, if that's what you mean :grinning:

dblommesteijn:
No I don't want it actually ;)

Issue #558, opened 2015-02-09
RestAPI naming convention JSON payload


dblommesteijn:
The RestAPI's naming convention for `GET /record/:id` [response-fields](https://github.com/EUDAT-B2SHARE/b2share/wiki/REST-API#response-fields) returns `recordID` (lower camelcase); whereas the other fields `open_access`, and `contact_email` use underscores to separate names. It seems wrong, why is this? Should we change it?

emanueldima:
The recordID is added manually and can be easily changed. The other field names are encoded in the database (marc fields 690__a) and more difficult to rename or map.

dblommesteijn:
I'm in favour of one single format, instead of two.

PR #672, opened 2015-06-11
Reject uploaded empty files via REST API

emanueldima:
This PR introduces some (minor) changes in the REST API results, potentially breaking existing clients (of which I presume are not many, if at all). These changes are:

 - uniform id names (e.g. `deposit_id` instead of `depositID`)
 - always return a JSON top level object, not an array (useful for security of JS clients)
 - renamed incorrect `Deposit` term with `record`

This PR fixes https://github.com/EUDAT-B2SHARE/b2share/issues/667 and fixes https://github.com/EUDAT-B2SHARE/b2share/issues/558

### DIFF
--- a/invenio/b2share/modules/b2deposit/restful.py
+++ b/invenio/b2share/modules/b2deposit/restful.py
@@ -113,7 +113,7 @@ def get_domain_metadata(domain_class, fieldset, bfo):
     return ret
 
 
-def get_record_details(recid, curr_user_email):
+def get_record_details(recid, curr_user_email=None):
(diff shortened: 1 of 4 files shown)


nharraud:
Nice! I had added a "FIXME: we should return IDs and not only location" in the tests I am currently adding. No need to commit it now ^^

nharraud:
We need to modify the WIKI page https://github.com/EUDAT-B2SHARE/b2share/wiki/REST-API once this is merged

emanueldima:
Yes, good point, please open a new issue.

PR #674, opened 2015-06-12
Setup travis to run b2share tests

### LINKED ISSUE #670: Make travis run (only) the B2SHARE tests
Currently the travis build does not include tests. As we started developing b2share specific tests, it is useful to include these in the build process.

emanueldima:
fixes #670


emanueldima:
The build failed once due to some bower connection error, but the tests seem to work well.

PR #767, opened 2015-08-26
Embargo

cjhak:


invenio/b2share/modules/b2deposit/b2share_marc_handler.py
```diff
@@ -157,6 +158,12 @@ def add_file_info(rec, form, email, sub_id, recid):
     else:
         fft_status = 'firerole: allow email "{0}"\ndeny all'.format(
             email)
+
+    if 'embargo_till' in form:
+        embargodate = parser.parse(form['embargo_till'])
+        embargodate = datetime.strftime(embargodate, '%Y-%m-%d')
+        fft_status = 'firerole: deny until "%s"\nallow any\n' % embargodate
```
emanueldima: As the first rule is `deny until ...` this will also deny the owner to access the record, and it's not what we want. I suppose the first rule must be `allow email ...` and then the deny rule.

emanueldima:
Let's not forget that this must also be ported to the rest api

cjhak:
Yes, I'll create an issue for that (REST API).

invenio/b2share/modules/b2deposit/b2share_marc_handler.py
cjhak: It seemed to me that `deny until` is equivalent to `allow email` until end of embargo. The way I wrote it seems to be how it has been done elsewhere.

I had to re-build my VM, so have not tested all use cases myself yet.

cjhak: @emanueldima Seems like you are right, `deny until`  does at least not behave as I had expected. However, I'm not sure what is optimal, so it is good to have it on the test server so we can get some feedback now. I'll try to squeeze in some improvements before the weekend.

PR #783, opened 2015-09-24
Add the new endpoints to the restapi draft

SarahBA:


b2share/modules/main/restful.py
```diff
@@ -0,0 +1,230 @@
...
+    def get(self, user_id, **kwargs):
+        """
+        Get a user profile information
+        """
+        return None
+
+
+def setup_app(app, api):
+    api.add_resource(CommunityList, '/api/communities')
+    api.add_resource(CommunityRes, '/api/communities/create')
```
emanueldima: We've decided not to use verbs in the URLs, so this one must go. The same functionality can be achieved by a POST in CommunityList.

b2share/modules/main/restful.py
```diff
@@ -0,0 +1,230 @@
...
+    api.add_resource(Community, '/api/communities/<int:community_id>')
+    api.add_resource(schemaList, '/api/schemas')
+    api.add_resource(schemaRes, '/api/schemas/create')
```
emanueldima: No verbs, see above.

b2share/modules/main/restful.py
```diff
@@ -0,0 +1,230 @@
...
+    api.add_resource(schema, '/api/schemas/<int:schema_id>')
+    api.add_resource(schemaDeprecate, '/api/schemas/<int:schema_id>/deprecate')
```
emanueldima: No verbs

b2share/modules/main/restful.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        Modify a community
+        """
+        return None
+
+
+# ----------------------------------------------- SCHEMA
+
+
+class schemaList(Resource):
```
emanueldima: Let's agree to start all class names with Capital Letters: https://www.python.org/dev/peps/pep-0008/#class-names

SarahBA:
Fixed the comments @emanueldima mentioned

PR #789, opened 2015-09-29
Serving ui frontend, add startup script, remove templates and views

emanueldima:


dblommesteijn:
There's an issue with Travis: 
```log
ERROR 1136 (21S01) at line 2: Column count doesn't match value count at row 1
The command "dbexec < ./install/_collections.sql" failed and exited with 1 during .
```

PR #802, opened 2015-11-04
alternative installation with docker-compse works again

httpPrincess:
adding additional dependencies in Dockerfile (both apt-get and pip)  rendered docker build successful.
fixes in docker-compose (old mysql password was wrongly interpreted).

All in all it makes alternative installation possible:
docker-compose build
docker-compose up

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,11 +38,15 @@ RUN apt-get update && \
     apt-get -qy install --fix-missing --no-install-recommends \
         gcc \
         git \
+        libcurl4-gnutls-dev \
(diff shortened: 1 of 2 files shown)


cjhak:
Branch 1.6 is the current maintenance branch and will be retired and succeeded by B2SHARE 2.0 in a few months time. We will not support Docker for 1.6. B2SHARE 2.0 is currently developed in the Evolution branch and already supports Docker. It will become the new master branch in the repository when it is mature enough.

I close this pull request.

PR #803, opened 2015-11-05
Use invenio 3 alpha modules

nharraud:
* Refactor new modules so that their api uses ContentNegotiatedMethodView instead of flask-restful Resource
* Use an environment variable to find the angular-js UI and use werkzeug SharedDataMiddleware to serve its files
* use invenio 3 alpha modules and in particular invenio-records-rest which provides the REST API for records. 

I did not refactor the travis script yet as I plan to make the UI more easily installable with pypi.


b2share/config.py
```diff
@@ -1,76 +1,24 @@
...
+from __future__ import absolute_import, print_function
+import os
 
-# Exclude all Invenio views
-PACKAGES_VIEWS_EXCLUDE = invenio_modules
-# Exclude all Invenio REST APIs
-PACKAGES_RESTFUL_EXCLUDE = invenio_modules
 
+# Identity function for string extraction
+def _(x):
```
emanueldima: _ is a terrible name for a function. I know it's used in invenio, but I still think it's a mistake. A function name must be greppable.

emanueldima:
I'd rather not remove the following folders:
````
b2share/modules/abuse
b2share/modules/datarequest
b2share/modules/b2deposit/b2share_model/metadata
b2share/modules/main/static/js
````
because they contain data that will be ported to the UI or DB, and we want their history until this happens.

I see you removed the webui module, how will the ui files be served?

b2share/config.py
nharraud: I will remove it for the time being but I'm not sure we will be able to do without it. It is in fact already grepped in Invenio as it is the one used to generate automatically the translation files.

nharraud:
Ok I'll remove the commit deleting the files then. We can do the cleaning later.

PR #805, opened 2015-11-06
One script to install and run b2share evolution

emanueldima:


devenv/start_here.sh
```diff
@@ -17,39 +12,73 @@
 export WORKON_HOME=~/.virtualenvs
```
nharraud: This does not seem to be used anymore. One can use `cdvirtualenv` or `$VIRTUAL_ENV` to go in the virtual environment directory.

emanueldima: Right

devenv/start_here.sh
```diff
@@ -17,39 +12,73 @@
...
+	workon $VIRTUALENV_NAME
+	cdvirtualenv && mkdir src
+fi
 
-# cdvirtualenv src/ui-frontend/
-# npm install
+docker-machine start $MACHINE_NAME
+if [ $? -ne 0 ]; then
+	echo; echo "### Create docker machine for b2share"
+	docker-machine create -d virtualbox $MACHINE_NAME
```
nharraud: this step is needed for mac only. We could detect the OS and use the local docker if we are on linux.
`DOCKER_IP` would then be `127.0.0.1`

emanueldima: Yes, and not only here, the location of the virtualenvwrapper.py is also hardcoded based on its osx location. I don't have any time to fix this though, I'd need to test it on a linux machine.

nharraud:
What about calling the script `run_demo.sh` rather `start_here.sh`? The current name seems to imply that it is required to run it and that there are other steps after it.

emanueldima:
Yes, done

devenv/run_demo.sh
```diff
@@ -0,0 +1,83 @@
...
+	git clone git@github.com:EUDAT-B2SHARE/ui-frontend.git
+	cd ui-frontend
+	npm install
+fi
+
+
+cdvirtualenv src
+if [ ! -d ./b2share ]; then
+	echo; echo "### Clone b2share"
+	git clone git@github.com:EUDAT-B2SHARE/b2share.git --branch evolution
```
nharraud: The issue here is that it will checkout the last version of the evolution branch. Not the currently checkouted version of b2Share. It means that the demo will fail when people will checkout the demo and then the script will checkout a failing version on Github. Same comment for the ui-frontend checkout.

emanueldima: Yes, we'll have to change it after we release. Do you have another idea?

nharraud:
Apart from the last comment, the script is working for me, both UI and API.

devenv/run_demo.sh
nharraud: The ui-frontend and b2share repo should be tagged with the same tag for beta released. So that we can know which versions are compatible.
We can change that once we have stable releases and don't need demo scripts.

nharraud: We could do this:
- check what tag is on the currently checkouted version
- create a symbolic link in the virtual environment
- clone the same tag from ui-frontend

if the current version is not tagged, checkout the last version of ui-frontend.

emanueldima: We can but it's probably overkill. Just checking a fixed tag is better.

nharraud: If you want. For later release we might just use docker so it will be simpler.

Issue #796, opened 2015-10-19
Re-direct B2SHARE users registered in B2SHARE to register in B2ACCESS upon login.


cjhak:
The possibility to log in directly to B2SHARE without using B2ACCESS should be removed. Existing registred users who try to login that way need to be re-directed to B2ACCESS to register anew and also informed that they should contact us (post to help-desk?) if they have material uploaded with their old accounts so the user accounts can be merged.

nharraud:
I don't think they need to inform us if they have material uploaded with their old account. The "migration" would be like that:
1/ user logs in with his old account. => the user sees a page which asks him to log in with B2Access
2/ the user logs into B2Access => we replace the old authentication with the new one for this user, keeping the same account.

What we need to know is if some users created both an account with B2Access and the old B2Share account system. Those accounts need to be merged.

I would simplify the message as "Please contact us if you don't seem to have access to all your records or for any other problem". If we start talking about multiple accounts people might get lost as they don't know how B2Share and B2Access work.

cjhak:
@nharraud I agree completely with your suggested solution.

nharraud:
I implemented something a little less agressive by letting users access their account with username and password even if they have connected a B2Access account. At each login the user is informed that the username/password mechanism will be disabled soon. A link is provided to connect the account to B2Access when it is not already done.

Issue #795, opened 2015-10-19
Remove/hide user registration


cjhak:
New users should register as such in B2ACCESS, not in B2SHARE. User registration in B2SHARE should be removed or hidden.

emanueldima:
Closed by #806

PR #806, opened 2015-11-12
global: register users with B2Access only

nharraud:
* Uses B2Access for new user registration. Encourage users using
  username/password mecanism to connect their B2Share account to
  a B2Access account.


nharraud:
Related to #795 #796

A flash message is displayed when the user login using username and password, encouraging him to connect his account to B2Access. The message contains a link enabling to connect the accounts. The username/password can still be used.
The message is less aggressive if the accounts are already connected. It then informs the user that the username password mechanism will be disabled soon.

nharraud:
The feature is working, I just leave it as 'In Progress' so that Emanuel can review it before it is merged.

emanueldima:
@nharraud Any idea why the build fails?

nharraud:
@emanueldima  Sadly no, I didn't have time to look into it yet. It's quite strange that it fails only in `devel` mode. I'll have some time this afternoon to search the problem's origin.

Once it is fixed, could you test it on your laptop? As this is mostly UI and we don't use selenium yet, there is no automatic test checking that it works. The modified pages are the login page, the main page once the login is done and the main page before login (the register link should send you to B2Access).

nharraud:
Ok this is just a dependency issue. The `devel` build uses master branches of some projects like `flask-admin` so that we can be compatible with their next version.
Apparently we fail with current flask-admin master branch. But as long as the release branch passes it should be ok.
Just to be sure I ran again the travis tests on a previously passing commit. It failed (https://travis-ci.org/EUDAT-B2SHARE/b2share/jobs/85269123)

emanueldima:
Yes, it's not the first time this happens. So far, Travis didn't help much in finding bugs but kept us from moving faster.

PR #820, opened 2016-01-19
B2share demo

nharraud:
@SarahBA @emanueldima This PR fixes the `takes no argument` issue, does some cleanup, adds a demo and a basic test testing the demo.

It also reenable travis tests. I mark the PR as `In Progress` as I don't know yet if the tests will succeed on travis.


emanueldima:
Why do we need a demo module?

nharraud:
@emanueldima for two reasons:
1/ Having a working demo is a good way to show the last features and have feedback. The demo module is in charge of adding all necessary data for this purpose.
2/ The demo will be the place where I move the generation of communities with their logo and everything related. As it will be possible to add communities dynamically, those resources should not be part of B2Share itself.

This demo could also be used later as a benchmarking tool for new instances. This would be similar to what Postgresql provides with `Pgbench`.

We might also move this demo outside of B2Share repository later on. I just though that it might be simpler at first and we can't forget to patch the demo when we refactor as its tests are
https://github.com/nharraud running on travis.

PR #821, opened 2016-01-21
global: coveralls service enabling

nharraud:
* Enables coveralls service in travis configuration.


nharraud:
@emanueldima This will enable us to detect when test coverage decreases :)

PR #824, opened 2016-01-22
Aalto template

llehtine:
First version of aalto metadata template


invenio/b2share/modules/b2deposit/b2share_model/metadata/aalto_metadata_config.py
```diff
@@ -0,0 +1,73 @@
...
+table_name = 'Aalto'
+image = 'domain-aalto.jpg'
+kind = 'project'
+domaindesc = 'Aalto University'
+
+# the domain administators can edit even published records
+admin_can_edit_published_record = True
+
+# only the domain administrators can deposit
+depositing_groups = ['aalto_domain_administrators']
```
emanueldima: If `depositing_groups` is defined, then only the domain administrators can deposit and nobody else. Is this the intended behaviour?

llehtine: I guess they only want the Aalto-people to be able to deposit to this domain. But isn't it possible to have normal users and administrators both on the group, and let the administrator handle membership? Although the administrator-term in the group name then might be misleading

emanueldima: No, in that case all the people would be able to edit all the aalto records. But we can make another depositing group, let's call it `aalto_domain_members`, different from the administrators group, and make that one the depositing group.

llehtine: Ok, that sounds good. So it would be like: depositing_groups = ['aalto_domain_administrators','aalto_domain_members'] ?

emanueldima: Yes, it should work that way.

PR #836, opened 2016-02-11
2. Refactored records module, ui improvements

emanueldima:


b2share/modules/records/restful.py
```diff
@@ -44,52 +59,64 @@ def community_to_json_serializer(data, code=200, headers=None):
...
-    view_name = 'record_list'
-
-    def __init__(self, *args, **kwargs):
-        super(RecordList, self).__init__(*args, **kwargs)
+class HttpResource(ContentNegotiatedMethodView):
+    def __init__(self, *args):
+        super(HttpResource, self).__init__(*args)
         self.serializers = {
-            'application/json': community_to_json_serializer,
+            'text/html': json_serializer,
```
nharraud: Why is the html media-type mapped to the json serializer?

emanueldima: I forgot about it, good catch :+1: I used it to more easily debug the api, using the browser directly.

PR #837, opened 2016-02-16
global: new community module implementation

nharraud:
* Refactors community module so that it uses current invenio 3
  modules.


nharraud:
@emanueldima ready for review. Missing features are:
- communities' schemas. They will be added later when the schemas module will be done.
- list of administrators. We still have to discuss the required access control granularity. Currently there are separate permissions for each type of actions (create, read, update, delete).

Programmatic and REST API are tested. Access control is only partially tested as we need to discuss it first.

.travis.yml
```diff
@@ -36,7 +36,8 @@ env:
   - REQUIREMENTS=devel REXTRAS=docs
 
 python:
-  - "2.7"
+  - "3.3"
```
emanueldima: Is there any value in testing older versions of Python? I'd like to have a faster travis response on pull requests.

b2share/modules/communities/api.py
```diff
@@ -18,73 +18,237 @@
...
+            date: True if the community is deleted, else False.
+        """
+        return self.model.deleted
+
+    @property
+    def name(self):
+        """Retrieve community's name."""
+        return self.model.name
+
+    @name.setter
```
emanueldima: Less is more. We don't need the setters.

b2share/modules/communities/config.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
emanueldima: If it's not used, it shouldn't exist, right?

.travis.yml
nharraud: @emanueldima Yes, There are some incompatibilities between those versions, and if we enable one day users to add plugins or customize their B2Share, they might use libraries which work only for older versions of python.
Also the overhead is not so big, most of the issues we had with Travis before where because it was working in 2.7 and not in 3.* or the contrary. It should happen a lot less now that we support only 3.*, thus testing locally will mean most of the time that it will work on Travis.

If for some reason there is a need to have very fast feedback from Travis, for example when we edit the `.travis.yml` file, it is still possible to comment the python versions while the PR is "in-progress" and uncomment them before merging. We just have to be extremely careful to not forget the uncommenting part and all reviewers for the `.travis.yml` should check it.

b2share/modules/communities/api.py
nharraud: My bad, I was sure I had removed them. I must have reverted the change during the rebases.

docs/make.bat
```diff
@@ -0,0 +1,263 @@
+@ECHO OFF
```
emanueldima: I'm thoroughly confused by this file. Shouldn't it have been a Makefile?

nharraud: A kind of makefile yes, but for windows. The docs directory is what you get by using `sphinx-quickstart`. It provides all the files needed to generate the sphinx documentation.

emanueldima: It should have created a Makefile also, if I read the sphinx docs correctly. How are we supposed to use a bat file?

nharraud: Yes you are right, I will redo the generation and update the PR.

nharraud: @emanueldima fixed. I also added the standard files like AUTHORS, and LICENSE.

I created a CONTRIBUTING.rst file as github automatically points to it when users create new issues. It is an enhanced version of what the wiki has. We can fix the wiki when the evolution branch replaces the master branch.

The `*.rst` files, including the README which I converted from markdown, are now included in the generated documentation.

This documentation can then be automatically published when B2Share will be released in Pypi.

emanueldima: Ok, I am happy now, but we must remove the .bat file. We shouldn't pretend that we support it.

PR #840, opened 2016-02-22
Eiscat template

llehtine:
eiscat template provided by Jani Heikkinen


invenio/b2share/modules/b2deposit/b2share_model/metadata/eiscat_metadata_config.py
```diff
@@ -0,0 +1,83 @@
...
+fields = [
+    {
+        'name': 'instrument',
+        'col_type': db.String(256),
+        'display_text': 'Instrument name',
+        'description': 'Instrument name',
+        'required': False
+    },
+    {
+        'name': 'kindat',
```
emanueldima: Do we have any meaningful information about this field? What is it? Is it the same as the one below, `kind_of_data_file`?

janiheikkinen: "'kindat' tells how the data have been processed. At EISCAT, we use two different processes marked as even/uneven (spectral/power analysis) values of kindat. These different profiles are often interleaved in the data."

emanueldima: Ok, I'll merge this request, but the whole purpose of the `description` field is to provide a meaningful description. Please consider adding one. It doesn't have to be long, and can refer to other online documentation.

PR #841, opened 2016-02-22
Communities rest adaptation, needed for ui

emanueldima:


b2share/modules/communities/api.py
```diff
@@ -85,10 +85,15 @@ def get(cls, id=None, name=None, with_deleted=False):
...
     @classmethod
+    # TODO: change this into a search function, not just a list of communities
     # TODO: a query should be given
     def get_all(cls, start, stop):
         """Searches for matching communities."""
-        raise NotImplementedError()
+        start = int(start)
+        stop = int(stop)
+        from .models import Community as CommunityMetadata
+        metadata = CommunityMetadata.query.limit(stop)[start:]
```
nharraud: Even if this is a temporary fix, you might want to have a default sorting, on 'created' for example.

b2share/modules/communities/views.py
```diff
@@ -127,7 +127,16 @@ def __init__(self, **kwargs):
...
     def get(self):
         """Retrieve a list of communities."""
-        raise NotImplementedError()
+        # TODO: change this to a search function, not just a list of communities
+        from .serializers import community_to_dict
+        start = request.args.get('start') or 0
+        stop = request.args.get('stop') or 20
+        community_list = Community.get_all(start, stop)
+        response = jsonify({
+            'communities': [community_to_dict(c) for c in community_list]
```
nharraud: Maybe we could agree on a common format for searching all resources. The default invenio-records-rest serializer generates this:
```python
{
  "hits": {
    "total": 42, # total number of results, not just on this "page"
    "hits": [
      {
        #record as serialized for the GET
      }
    ]
  },
 "links": {
   "next": <URL>,
   "prev": <URL>
 }
}
```
I didn't add here the aggregations as we can add them later.

Also this serialization can be changed if we need something else. It is not hard coded in invenio-records-rest.

nharraud:
Otherwise it looks good :+1:

b2share/modules/communities/views.py
emanueldima: ok, made an issue for this one
https://github.com/EUDAT-B2SHARE/b2share/issues/854

PR #865, opened 2016-03-03
Added EISCAT

janiheikkinen:


emanueldima:
Correct, but please also add the EISCAT to the configuration file in the installation script: https://github.com/EUDAT-B2SHARE/b2share/blob/1.7/install/install_b2share.sh#L102

@llehtine Lassi, if you hadn't already, you should also add EISCAT to the CFG_B2SHARE_DOMAINS on production/training

llehtine:
Hi, 

It's defined on both training/production. 

Br, 
Lassi

PR #867, opened 2016-03-03
schemas: new B2Share Schemas module implementation

nharraud:
* NEW Adds Block and Root Schemas API.
* NEW Adds Community Schemas API.


nharraud:
@emanueldima what do you think of this design?
I just did a prototype for the programatic API and the database model.
There is a test showing how it can be used.
There is no REST API. We can discuss it IRL if you want.

nharraud:
@tiborsimko You were interested in the B2Share Schema module design, here is the PR. As said to @emanueldima this is only a rapid prototype to test a design.

nharraud:
@emanueldima can you please take a look at this version?
It enables communities to directly submit json-schemas as you suggested.
The REST API enables only GET operations for the time being.
If you are OK we could merge this so that you can continue your work on the UI. The rest of the REST API can be implemented later.

demo/b2share_demo/cli.py
```diff
@@ -37,22 +52,161 @@ def demo():
 @click.option('-v', '--verbose', is_flag=True, default=False)
 def load(verbose):
     """Load demonstration data."""
-    _create_records()
+    with db.session.begin_nested():
+        communities = _create_communities(verbose)
+        block_schemas = _create_block_schemas(communities)
```
emanueldima: But `_create_block_schemas` doesn't return anything.

demo/b2share_demo/cli.py
```diff
@@ -37,22 +52,161 @@ def demo():
...
+                    'block_schemas'].items():
+                block_schema = BlockSchema.create_block_schema(
+                    community.ref.id,
+                    schema_name)
+                for json_schema in schema['versions']:
+                    block_schema.create_version(json_schema)
+                nb_block_schemas += 1
+    click.secho('Created {} block schemas!'.format(nb_block_schemas),
+                fg='green')
+
```
emanueldima: `return` expected?

demo/b2share_demo/cli.py
nharraud: Good catch, I'll have to remove the assignment.
I plan to split this in multiple commands later. They won't depend on other functions' return value but access the database instead.

demo/b2share_demo/cli.py
nharraud: No, the above assignment has to be removed. I missed it after some refactoring.

PR #888, opened 2016-03-22
Indexing new records

emanueldima:


b2share/modules/records/triggers.py
```diff
@@ -36,6 +37,7 @@
 
 def register_triggers(app):
     before_record_insert.connect(set_record_schema)
+    before_record_insert.connect(index_record)
```
nharraud: Records are not yet validated when `before_record_insert` is fired. Can you use `after_record_insert` instead?

b2share/modules/records/triggers.py
```diff
@@ -48,3 +50,14 @@ def set_record_schema(record, **kwargs):
...
     schema = CommunitySchema.get_community_schema(community_id)
     record['$schema'] = community_schema_json_schema_link(schema)
+
+
+def index_record(record, **kwargs):
+    current_search_client.index(
+        index='records',
+        doc_type='record',
+        id=record.id,
+        body=record,
```
nharraud: Later we might also need to index *internal* metadata fields like `created` and `updated`.

PR #889, opened 2016-03-22
Run demo updates

emanueldima:


devenv/run_demo.sh
```diff
@@ -67,6 +77,7 @@ export B2SHARE_BROKER_URL="redis://${DOCKER_IP}:6379/0"
 export B2SHARE_CELERY_RESULT_BACKEND="redis://${DOCKER_IP}:6379/1"
 export B2SHARE_SECRET_KEY=$(base64 /dev/urandom | tr -d '/+' | dd bs=32 count=1 2>/dev/null)
 export SEARCH_ELASTIC_HOSTS="${DOCKER_IP}:9200"
+export B2SHARE_SERVER_NAME="localhost:5000"
```
nharraud: `b2share schemas init` on line 62 needs `B2SHARE_SERVER_NAME`.

nharraud:
We can merge this PR once the `B2SHARE_SERVER_NAME` initialization is moved upper in the script.

Later I would like to make the script safer. As we don't run the script without `/bin/bash -e` option it continues running even when an error occurs.
I would also prefer if it got simpler as it plays too much with virtual environments. I had to comment part of the script in order to test it as I didn't want it to interfer with my already existing virtual environment and it might have polluted my global python environment if it failed.

emanueldima:
Ok, please check now. Refactoring the script is a good idea :+1:

PR #891, opened 2016-04-12
global: invenio integration improvement

nharraud:
* NEW Adds Files support by integrating invenio-files-rest.

* NEW Adds triggers for record indexation in Elasticsearch.

* NEW Uses records' UUID as Persistent Identifier instead of
  auto-incremented integer.

* Refactors tests so that they use the flask application as created
  by b2share application factory.


nharraud:
@emanueldima it is far from being perfect but it enables some of the features required for the demo.
- The file support is not complete as the references are not added to the record. But it is possible to upload and get files.
- The indexing is hard-coded. I plan to replace it with invenio-indexer later. This will enable bulk reindexing when needed.

PR #920, opened 2016-04-18
Demo fix

nharraud:
* Fixes the docker demo

### DIFF
--- a/dockerize/Dockerfile
+++ b/dockerize/Dockerfile
@@ -26,7 +26,10 @@ RUN npm install && \
     node_modules/webpack/bin/webpack.js -p
 
 WORKDIR /eudat/b2share
-RUN pip install -r requirements.txt
+RUN pip install requirements-builder
+RUN requirements-builder --level=min setup.py > lowest-requirements.txt
+RUN pip install -r lowest-requirements.txt
+RUN pip install -e .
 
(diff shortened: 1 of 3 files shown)


nharraud:
@emanueldima could you test this?

It does not however fix the *development environment* install as we should be on bleeding edge for this one in order to detect and fix incompatibilities.

nharraud:
And only *lowest* variant of travis tests will pass for this PR as the other variants use more recent and not backward compatible versions of Invenio 3.

emanueldima:
The container builds and run correctly, but any operation with records fails with HTTP 500. Any idea why?

nharraud:
Strange. I just tested the docker container with GET and PATCH and both are working properly
`curl -XGET 'http://192.168.99.100:5000/api/records/a1c2ef96-a1e4-46fa-9bd7-a2a46d2242d4'`

`curl -v -XPATCH 'http://192.168.99.100:5000/api/records/a1c2ef96-a1e4-46fa-9bd7-a2a46d2242d4' -H "Content-Type:application/json-patch+json" -H "Accept: */*" -d '[{ "op": "replace", "path": "/title", "value": "patch title" }]'`

emanueldima:
Yes, getting an individual record works, but getting /api/records fails now, probably because of the query module.

PR #941, opened 2016-05-04
global: fix invenio dependencies

nharraud:
* Fixes Invenio dependencies after some non backward compatible
  refactoring.


nharraud:
@emanueldima This forces the lowest version of invenio-records-rest to 1.0.0a13 as it fixes the issue we had.

PR #945, opened 2016-05-10
Update lter_metadata_config.py

llehtine:


emanueldima:
@llehtine I would like to know what this URL is for. If it's pointing to additional metadata, then why isn't this metadata together with the data, in B2SHARE? If it's a schema, then the description should provide more detail. Everyone browsing through the community fields should understand what the fields are about.

llehtine:
Hi, this was requested by user. I've send the details to you by email

Issue #915, opened 2016-04-18
v2: HTTP 500 when updating (PUTting) a record


emanueldima:
It's due to the lack of an _internal field in the metadata sent by the user.

emanueldima:
closed by https://github.com/EUDAT-B2SHARE/b2share/issues/948

PR #948, opened 2016-05-14
Add record owner, more tests, fix http put

emanueldima:
- add record owner, with demo updates
- index records on update
- fix http put (_internal field)
- more HTTP record tests: put/patch/get 

closes https://github.com/EUDAT-B2SHARE/b2share/issues/915

### DIFF
--- a/b2share/modules/records/triggers.py
+++ b/b2share/modules/records/triggers.py
@@ -26,9 +26,14 @@
 from __future__ import absolute_import, print_function
 
 import uuid
-
 import pytz
-from invenio_records.signals import before_record_insert, after_record_insert
+
+from flask import abort
+from flask_login import current_user
+
+from invenio_records.signals import (
+        before_record_insert, after_record_insert,
+        before_record_update, after_record_update)
 
 from b2share.modules.schemas.api import CommunitySchema
 from b2share.modules.schemas.serializers import \
@@ -37,14 +42,20 @@ from invenio_files_rest.models import Bucket
 from invenio_search import current_search_client
 
 from .errors import InvalidRecordError
-from .constants import (RECORDS_INTERNAL_FIELD as interal_field,
+from .constants import (RECORDS_INTERNAL_FIELD as internal_field,
                         RECORDS_BUCKETS_FIELD as buckets_field)
 
 
 def register_triggers(app):
     before_record_insert.connect(set_record_schema)
+    before_record_insert.connect(set_record_owner)
     before_record_insert.connect(create_record_files_bucket)
     after_record_insert.connect(index_record)
+
+    before_record_update.connect(check_record_immutable_fields)
+    before_record_update.connect(re_add_record_files_bucket)
+    after_record_update.connect(index_record)
+
 
 
 def set_record_schema(record, **kwargs):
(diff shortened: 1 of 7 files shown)


b2share/modules/records/triggers.py
```diff
@@ -60,12 +70,18 @@ def set_record_schema(record, **kwargs):
     record['$schema'] = community_schema_json_schema_link(schema)
 
 
+def set_record_owner(record, **kwargs):
+    if not current_user.is_authenticated:
```
nharraud: It think that it would be better to leave the `is_authenticated` check to the permission factory. See https://github.com/inveniosoftware/invenio-records-rest/blob/144dee816ee05c10e09e72eaf2f5b18c74285977/invenio_records_rest/views.py#L264

The permission factory just has to return a permission which returns False when `can()` is called.

b2share/modules/records/triggers.py
```diff
@@ -82,3 +98,17 @@ def index_record(record):
         version=record.revision_id,
         version_type='external_gte',
     )
+
+
+def check_record_immutable_fields(record):
```
nharraud: It might be better to check this sooner. Either in the loaders: https://github.com/inveniosoftware/invenio-records-rest/blob/144dee816ee05c10e09e72eaf2f5b18c74285977/invenio_records_rest/views.py#L537 or in the permission factory.

tests/b2share_functional_tests/test_submission.py
```diff
@@ -47,122 +48,227 @@
...
+    assert get_bucket_links(record_put_res.headers) == bucket_links
+    _test_get_record(client, record_url, record_put_data, bucket_links)
+
+
+def _test_file_ops(client, bucket_links, count):
+    for num in range(count):
+        # Test file upload
+        headers = {'Accept': '*/*'}
+        object_url = '{0}/{1}'.format(bucket_links[0], 'myfile.dat')
+        content = 'contents2~`!+üøåé+%20+ +\n{}'.format('word ' * num)
```
nharraud: Sorry if I missed something but I don't see what the reason for the cryptic string.

nharraud:
@emanueldima The refactoring of the test looks good. Concerning authentication check and user input validation such as immutable fields, I would rather not use signals for those as Invenio already provides other means which "look" more clean to me. This is a subjective point of view.

I think we can
- check that the user is authenticated in the permission factories.
- set the owner in the input loader.

For the input validation, such as immutable fields, we could do it in the permission factory. The issue with this solution is that we load twice the input. Another way to do it would be to check the additional permissions needed in the loader, i.e: the loader would load the input, then see what fields are modified and deduce the permissions needed (ex: change_owner_need), then it would check that the user is allowed to do this action. If a modification is illegal for everybody (ex: the resulting record does not match the JSON Schema anymore) it returns 404.

b2share/modules/records/triggers.py
emanueldima: I don't disagree, but I want to make sure in this function that user.id exists and is valid. More of a double check. If the permission check comes first, it's even better. I will change this to abort(401) as you said.

b2share/modules/records/triggers.py
emanueldima: You're right, these should be permissions. I will remove them once we have that module; or could you remove them in the same commit?

b2share/modules/records/triggers.py
nharraud: Ok we can keep it like that for now. What about moving the owner-update code to the user input loader as suggested in my last comment? It would allow more flexibility for use cases such as the creation of already existing records by the admin (ex: when loading the data from B2Share 1.7 in 2.0).

b2share/modules/records/triggers.py
nharraud: I'll see when the permissions commit is ready. If the change is not too big I'll include it.

tests/b2share_functional_tests/test_submission.py
emanueldima: No special purpose, I just wanted to have unicode text of a variable length.

b2share/modules/records/triggers.py
emanueldima: Ok, but then shouldn't all the `before_` triggers be moved to loaders?

nharraud: I think you are right for the `$schema` part. Regarding the bucket I don't know. I have to check if there is any issue with the loader creating other resources such as buckets.

emanueldima: https://github.com/EUDAT-B2SHARE/b2share/issues/955

PR #956, opened 2016-05-18
Proper HTTP errors when record validation fails

### LINKED ISSUE #911: Useful error message when posting/patching invalid record metadata
Currently when a record validation fails the exception is simply caught by flask and results in a HTTP 500 Internal Server Error. We want instead a HTTP 400 error message with some simple explanation.

emanueldima:
close #911


b2share/modules/records/errors.py
```diff
@@ -36,3 +41,16 @@ class AlteredRecordError(B2ShareRecordsError):
     """Raise when a record update changes what is considered
        immutable record data."""
 
+
+def register_error_handlers(app):
+    @app.errorhandler(ValidationError)
+    def handle_ValidationError(err):
```
nharraud: I would rather keep to PEP8 naming style when possible (https://www.python.org/dev/peps/pep-0008/#function-names), i.e: `handle_validation_error`

emanueldima:
@nharraud Any other issues?

b2share/modules/records/triggers.py
```diff
@@ -52,7 +53,12 @@ def register_triggers(app):
     before_record_insert.connect(create_record_files_bucket)
     after_record_insert.connect(index_record)
 
+    # TODO(edima): replace this check with explicit permissions
     before_record_update.connect(check_record_immutable_fields)
+
+    # don't trust the user, reset record schema
+    #   from (possibly updated) community field
+    before_record_update.connect(set_record_schema)
```
nharraud: `set_record_schema ` sets the schema to the last version of a community schema. Thus setting it again would override an old schema version with a new one, potentially making the record invalid.

b2share/modules/records/errors.py
```diff
@@ -36,3 +41,16 @@ class AlteredRecordError(B2ShareRecordsError):
...
 
+
+def register_error_handlers(app):
+    @app.errorhandler(ValidationError)
+    def handle_validation_error(err):
+        field = '/'.join(err.path)
+        if err.validator == 'required' or err.validator == 'additionalProperties':
+            try:
+                field = err.message.split('\'')[1]
+            except Exception:
```
nharraud: Catching silently all exceptions might hide some nasty bugs. Why not just catch `IndexError` in case the string is not as expected and the split returns an array with len < 2?

b2share/modules/records/triggers.py
```diff
@@ -52,21 +53,28 @@ def register_triggers(app):
     before_record_insert.connect(create_record_files_bucket)
     after_record_insert.connect(index_record)
 
+    # TODO(edima): replace this check with explicit permissions
     before_record_update.connect(check_record_immutable_fields)
-    before_record_update.connect(re_add_record_files_bucket)
+
+    # don't trust the user, reset record schema on update
```
emanueldima: Updated

PR #963, opened 2016-06-10
Added nginx for docker-compose and config-file creation script

llehtine:


dockerize/create_nginx_conf.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+echo "What is your servers FQDN? "
+read FQDN
+echo "What is your servers IP? "
+read IP_ADDR
+
+cat <<EOF > nginx/b2share.conf
+server {
+        listen 80;
```
emanueldima: We can not allow unencrypted traffic. Sometimes the web UI sends session confidential information to the server.

llehtine: Ok. I will have to remove the normal 80-interface from the config.

llehtine: Although, the OAI-standard does require unencrypted http

emanueldima: You're right, port 80 should be open for /oai2d only.

PR #964, opened 2016-06-10
global: invenio-deposit integration

nharraud:
* NEW Integrates invenio-deposit. This changes the submission
  workflow.

WARNING: functional tests have been disabled or replaced as the way records are created changed and there is no time to FIX them now.


certs/g_cacert.pem
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
emanueldima: What is this one for?

nharraud: Oups, my bad. I created this file when I was retrieving public certificates to make the B2Access test instance SSL work. Nothing confidential, but it should be removed. (Reminder to myself: sending PR past 10PM is a bad idea)

PR #982, opened 2016-06-29
Oai-pmh integration

emanueldima:
After startup the oai-pmh endpoint is available at `/api/oai2d`:

http://localhost:5000/api/oai2d?verb=ListRecords&metadataPrefix=oai_dc 

The OAI-PMH functionality depends on this PR: https://github.com/inveniosoftware/invenio-oaiserver/pull/73


b2share/modules/records/triggers.py
```diff
@@ -94,6 +90,11 @@ def set_record_owner(record, **kwargs):
     record['owner'] = str(current_user.id)
 
 
+def add_oai_id(record):
```
jirikuncar: What about using PID minter (https://github.com/inveniosoftware/invenio-oaiserver/blob/master/invenio_oaiserver/minters.py) or defining your own? Each OAIID should be registered.

emanueldima: You're right, of course, I will add it.

nharraud: @emanueldima I see that `OAIIDProvider` and `oaiid_fetcher ` [are used in multiple places in invenio-oaiserver](https://github.com/inveniosoftware/invenio-oaiserver/blob/d68f62741c422b34a1438e53b0694570f9ac8e6b/invenio_oaiserver/response.py#L265). Thus, as @jirikuncar said, we need to register the PID. We could add this minter:
```python
def b2share_oaiid_minter(record_uuid, data):
    """Mint record identifiers."""
    pid_value = data.get('_oai', {}).get('id')
    if pid_value is None:
        assert '_deposit' in data and 'id' in data['_deposit']
        pid_value = current_app.config.get('OAISERVER_ID_PREFIX', '') + str(
            data['_deposit']['id']
        )
    provider = OAIIDProvider.create(
        object_type='rec', object_uuid=record_uuid,
        pid_value=str(pid_value)
    )
    data.setdefault('_oai', {})
    data['_oai']['id'] = provider.pid.pid_value
    return provider.pid
```
And use it for every created record, which means: when the deposit is published for the first time. The minter would then be called by the `b2share.modules.deposit.api.Deposit.publish` method (in PR #983)

Note that here I am using the deposit ID as we have the same value for the published record. However at some point we have to check if we can use the same ID for EPIC PIDs too, and if not, decide if we use the internal ID or the EPIC PID as the OAI ID.

Note for later documentation: we have to show organizations deploying B2Share how to configure `OAISERVER_ID_PREFIX`. It must contain the organization namespace as described here: http://www.openarchives.org/OAI/2.0/guidelines-oai-identifier.htm

```
Organizations must choose namespace-identifier values which correspond to a domain-name that they have registered, and are committed to maintaining.
```

emanueldima: Better to wait for https://github.com/EUDAT-B2SHARE/b2share/pull/983 to be ready then, before merging this one.

emanueldima:
@nharraud @jirikuncar I've added the oai minter, please have a look

setup.py
```diff
@@ -31,14 +31,19 @@
...
     'invenio-accounts>=v1.0.0a12,<1.1.0',
     'invenio-base>=1.0.0a11,<1.1.0',
     'invenio-celery>=1.0.0a4,<1.1.0',
     'invenio-config>=1.0.0a1,<1.1.0',
     'invenio-db>=1.0.0a9,<1.1.0',
+    'invenio-deposit>=1.0.0a1,<1.1.0',
     'invenio-files-rest>=1.0.0a4,<1.1.0',
     'invenio-mail>=1.0.0a4,<1.1.0',
+    'invenio-marc21>=1.0.0a2',
+    'invenio-oaiserver>=1.0.0a5,<1.1.0', # a7 really
```
nharraud: why the comment about a7? I see only versions up to a5.

nharraud:
Looks good. I searched why the tests are not passing. A new release of `invenio-oauthclient` should fix the issue, I created the PR for that inveniosoftware/invenio-oauthclient#71. Once it is merged and released we should update the minimum version for `invenio-oauthclient`.

nharraud:
@emanueldima thanks. Just the fix we needed :)

PR #983, opened 2016-07-01
global: deposit publication workflow support

nharraud:
* NEW Adds Deposit publication workflow and support it with programmatic
  and REST API.

* Cleans unused code.

* Lets invenio-deposit set the deposit owner. Remove unused trigger.

### DIFF
--- a/b2share/modules/deposit/api.py
+++ b/b2share/modules/deposit/api.py
@@ -24,19 +24,29 @@
 """B2Share Deposit API."""
 
 import uuid
+from enum import Enum
 
-from flask import current_app, url_for
+from flask import url_for
 from invenio_deposit.api import Deposit as DepositRecord
 from b2share.modules.schemas.api import CommunitySchema
-from b2share.modules.schemas.serializers import \
-    community_schema_json_schema_link
 from b2share.modules.deposit.minters import b2share_deposit_uuid_minter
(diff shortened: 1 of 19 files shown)


nharraud:
There is a coverage regression as I removed many previously covered lines.

b2share/modules/deposit/api.py
```diff
@@ -74,6 +81,9 @@ def build_deposit_schema(self, record):
     @classmethod
     def create(cls, data, id_=None):
         """Create a deposit."""
+        # check that the status field is not set
+        assert 'publication_state' not in data
```
emanueldima: Why not ` raise InvalidDepositDataError('"publication_state" field should not be set.')`?

nharraud: Because we have two inputs for the programmatic API.
1/ CLI: It should use the API properly whatever happens.
2/ REST API: it should filter its input using the input loaders. The loaders should be the one raising the error.

Thus if we arrive here with a `publication_state` set there is something wrong going on,
Later we might disable this check as we will have to handle B2Share migration (2.0->2.x).

nharraud: But yes we can also raise an exception instead. The would should be similar, i.e fail on bad input.

emanueldima: Yes, so my comment is on the fact that assert will raise a generic exception, and it's better to have something more meaningful.

PR #986, opened 2016-07-06
global: addition of REST api for Block Schemas

PaulinaLach:
* NEW Adds a new:
  - POST - creates new Block Schema
  - PUT - creates new Block Schema Version for Block Schema
  - GET - gets list of schemas

Signed-off-by: PaulinaLach <paulina.malgorzata.lach@cern.ch>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 .settings
 .version
 .ropeproject
+.idea/
 Invenio.egg-info
 Makefile.in
 TAGS
--- a/b2share/modules/schemas/api.py
+++ b/b2share/modules/schemas/api.py
@@ -33,12 +33,14 @@ from invenio_db import db
 from sqlalchemy.orm.exc import NoResultFound
 
 from b2share.modules.communities import Community
+from b2share.modules.schemas.helpers import validate_json_schema
 
+from jsonpatch import apply_patch
 from .errors import BlockSchemaDoesNotExistError, BlockSchemaIsDeprecated, \
     CommunitySchemaDoesNotExistError, InvalidBlockSchemaError, \
     InvalidJSONSchemaError, InvalidRootSchemaError, \
-    RootSchemaDoesNotExistError
-from .helpers import validate_json_schema
+    RootSchemaDoesNotExistError, InvalidSchemaVersionError,\
+    SchemaVersionExistsError
 
 
 class RootSchema(object):
@@ -169,7 +171,7 @@ class BlockSchema(object):
                 filters.append(BlockSchemaModel.name == name)
 
             return [cls(model) for model in BlockSchemaModel.query.filter(
-                *filters).all()]
+                *filters).order_by(BlockSchemaModel.created).all()]
         except NoResultFound as e:
             raise BlockSchemaDoesNotExistError() from e
 
@@ -194,7 +196,52 @@ class BlockSchema(object):
             raise InvalidBlockSchemaError() from e
         return block_schema
 
-    def create_version(self, json_schema):
+    def update(self, data, clear_fields=False):
+        """Update multiple fields of the schema at the same time.
+        If the update fails, none of the schema's fields are modified.
+        Args:
+            data (dict): it replaces the given values.
+            clear_fields (bool): if True, set not specified fields to None.
+        Returns:
+            :class:`BlockSchema`: self
+        Raises:
+            b2share.modules.schemas.InvalidBlockSchemaError: The
+                schema update failed because the resulting schema is
+                not valid.
+        """
+        try:
+            with db.session.begin_nested():
+                if clear_fields:
+                    for field in ['name']:
+                        setattr(self.model, field, data.get(field, None))
+                else:
+                    for key, value in data.items():
+                        setattr(self.model, key, value)
+                db.session.merge(self.model)
+        except sqlalchemy.exc.IntegrityError as e:
+            raise InvalidBlockSchemaError() from e
+        return self
+
+    def patch(self, patch):
+        """Update the schema's metadata with a json-patch.
+        Args:
+            patch (dict): json-patch which can modify the following fields:
+                name, description, logo.
+        Returns:
+            :class:`BlockSchema`: self
+        Raises:
+            jsonpatch.JsonPatchConflict: the json patch conflicts on the
+                block schema.
+            jsonpatch.InvalidJsonPatch: the json patch is invalid.
+            b2share.modules.block_schemas.errors.InvalidBlockSchemaError: The
+                block schema patch failed because the resulting block schema is
+                not valid.
+        """
+        data = apply_patch({'name': self.model.name}, patch, True)
+        self.update(data)
+        return self
+
+    def create_version(self, json_schema, version_number=None):
         """Create a new version of this schema and release it.
 
         Args:
@@ -224,8 +271,14 @@ class BlockSchema(object):
             ).filter(
                 BlockSchemaVersionModel.block_schema == self.model.id
             ).one().last_version
+
             new_version = (last_version + 1 if last_version is not None
                            else 0)
+            if version_number and new_version < version_number:
+                raise InvalidSchemaVersionError(last_version)
+            elif version_number and new_version > version_number:
+                raise SchemaVersionExistsError(last_version)
+
             model = BlockSchemaVersionModel(
                 block_schema=self.model.id,
                 # we specify the separators in order to avoir whitespaces
@@ -310,6 +363,11 @@ class BlockSchema(object):
                 An iterator on released block schema versions.
         """  # noqa
         return BlockSchemaVersionsIterator(self)
+
+    @property
+    def updated(self):
+        """Retrieve the updated field value of this Block Schema."""
+        return self.model.updated
 
 
 class BlockSchemaVersionsIterator(object):
--- a/b2share/modules/schemas/errors.py
+++ b/b2share/modules/schemas/errors.py
@@ -82,6 +82,42 @@ class BlockSchemaVersionIsReleased(B2ShareSchemasError):
     """Exception raised while trying to modify a released block schema."""
     pass
 
+
+#
+# BLOCK SCHEMA VERSION ERRORS
+#
+
+class InvalidSchemaVersionError(B2ShareSchemasError):
+    """Exception raised when trying to add a version with wrong id."""
+
+    MESSAGE = """Version number is invalid. Provided version number should
+                follow the last existing version number, which currently
+                is {0}."""
+
+    def __init__(self, last_version):
+        """Constructor."""
+        self.last_version = last_version
+        """Error message."""
+        super(InvalidSchemaVersionError, self).__init__(
+            self.MESSAGE.format(last_version)
+        )
+
+
+class SchemaVersionExistsError(B2ShareSchemasError):
+    """Exception raised when trying to add an existing version."""
+
+    MESSAGE = """Version number already exists. Provided version number should
+                follow the last existing version number, which currently
+                is {0}."""
+
+    def __init__(self, last_version):
+        """Constructor."""
+        self.last_version = last_version
+        """Error message."""
+        super(SchemaVersionExistsError, self).__init__(
+            self.MESSAGE.format(last_version)
+        )
+
 #
 # COMMUNITY SCHEMA VERSION ERRORS
 #
--- a/b2share/modules/schemas/serializers.py
+++ b/b2share/modules/schemas/serializers.py
@@ -55,6 +55,16 @@ def block_schema_version_json_schema_link(block_schema_version, **kwargs):
 
 
 def block_schema_version_to_dict(block_schema_version):
+    """Serializes block schema version to dict.
+
+    Args:
+        block_schema_version
+            (:class:`b2share.modules.schemas.api:BlockSchemaVersion`):
+            block schema version that will be serialized.
+
+    Returns:
+        dict: serialized BlockSchemaVersion.
+    """
     return dict(
         id=block_schema_version.block_schema.id,
         version=block_schema_version.version,
@@ -64,6 +74,20 @@ def block_schema_version_to_dict(block_schema_version):
 
 def block_schema_version_to_json_serializer(block_schema_version, code=200,
                                             headers=None):
+    """Serializes block schema version to json response.
+
+    Args:
+        block_schema_version
+            (:class:`b2share.modules.schemas.api:BlockSchemaVersion`):
+            block schema version that will be serialized.
+
+        code: http response status.
+
+        headers: additional http response headers.
+
+    Returns:
+        Response: serialized response from BlockSchemaVersion.
+    """
     response = jsonify(block_schema_version_to_dict(block_schema_version))
     response.status_code = code
     if headers is not None:
@@ -80,7 +104,7 @@ def community_schema_self_link(community_schema, **kwargs):
 
         **kwargs: additional parameters given to flask.url_for.
 
-    :Returns:
+    Returns:
         str: link pointing to the given community schema.
     """  # noqa
     return url_for(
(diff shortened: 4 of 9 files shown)


b2share/modules/schemas/views.py
```diff
@@ -139,12 +147,79 @@ def get(self, community_schema):
...
 
 blueprint.add_url_rule(
     '/communities/<string:community_id>/schemas/<string:schema_version_nb>',
     view_func=CommunitySchemaResource
-    .as_view(CommunitySchemaResource.view_name))
+        .as_view(CommunitySchemaResource.view_name))
+
+
+blueprint.add_url_rule(
+    '/api/schemas',
```
emanueldima: I think this string should be only `'/schemas'`, without the `/api`

.gitignore
```diff
@@ -22,6 +22,7 @@
 .settings
 .version
 .ropeproject
+.idea/vcs.xml
```
nharraud: Could you ignore `.idea` and not `.idea/vcs.xml`

b2share/modules/schemas/serializers.py
```diff
@@ -60,7 +60,10 @@ def block_schema_version_to_dict(block_schema_version):
 
 def block_schema_version_to_json_serializer(block_schema_version, code=200,
                                             headers=None):
-    response = jsonify(block_schema_version_to_dict(block_schema_version))
+    if block_schema_version:
```
nharraud: Is this what you meant?
```python
if block_schema_version is not None:
```
`None` is not the only value which casts to `False` in python.

b2share/modules/schemas/serializers.py
```diff
@@ -110,3 +113,41 @@ def community_schema_to_json_serializer(community_schema, code=200,
     if headers is not None:
         response.headers.extend(headers)
     return response
+
+
+def new_schema_to_json_serializer(schema, code=200,
+                                            headers=None):
+    if schema:
```
nharraud: same here, use `is not None` if this is what you meant.

b2share/modules/schemas/serializers.py
```diff
@@ -110,3 +113,41 @@ def community_schema_to_json_serializer(community_schema, code=200,
...
+        response = jsonify(block_schema_to_dict(schema))
+    else:
+        response = jsonify({})
+    response.status_code = code
+    if headers is not None:
+        response.headers.extend(headers)
+    return response
+
+
+def block_schema_to_dict(schema):
```
nharraud: Please document functions even when they are simple.

b2share/modules/schemas/views.py
```diff
@@ -139,12 +147,79 @@ def get(self, community_schema):
         return community_schema
 
 
+class BlockSchemaResource(ContentNegotiatedMethodView):
```
nharraud: This needs to be split into two resource classes. BlockSchemaResource and BlockSchemaListResource

b2share/modules/schemas/views.py
```diff
@@ -139,12 +147,79 @@ def get(self, community_schema):
...
+    view_name = 'block_schema_creating_item'
```
nharraud: We should follow the endpoint naming of the other Invenio modules. Here it would be `block_schema_list `

tests/b2share_unit_tests/schemas/test_schemas_rest.py
```diff
@@ -68,3 +69,72 @@ def test_valid_get(app, test_communities):
...
+                       ('Accept', 'application/json')]
+            res = client.get(
+                url_for('b2share_schemas.block_schema_creating_item'),
+                data=json.dumps({
+                    'community_id': community_id
+                }),
+                headers=headers)
+            assert res.status_code == 200
+            response_data = json.loads(res.get_data(as_text=True))
+            assert isinstance(response_data['schemas'], list)
```
nharraud: The content of the schemas list is not tested.

tests/b2share_unit_tests/schemas/test_schemas_rest.py
```diff
@@ -68,3 +69,72 @@ def test_valid_get(app, test_communities):
...
+                data=json.dumps({
+                    'name': 'abc',
+                    'community_id': community_id
+                }),
+                headers=headers)
+            assert res.status_code == 201
+            response_data = json.loads(res.get_data(as_text=True))
+            assert response_data['name'] == BlockSchema.get_block_schema(
+                response_data['schema_id']
+            ).name
```
nharraud: `community_id` is not tested.

b2share/modules/schemas/views.py
```diff
@@ -139,12 +147,96 @@ def get(self, community_schema):
...
+        .as_view(CommunitySchemaResource.view_name))
+
+
+blueprint.add_url_rule(
+    '/schemas/<string:schema_id>',
+    view_func=BlockSchemaResource
+        .as_view(BlockSchemaResource.view_name))
+
+blueprint.add_url_rule(
+    '/schemas/<string:community_id>',
```
nharraud: the endpoint should be `/schemas/.
For POST the community_id should be in the data.
for GET it should be given as a query parameter.

b2share/modules/schemas/views.py
```diff
@@ -139,12 +147,96 @@ def get(self, community_schema):
...
+                    'application/json': block_schema_version_to_json_serializer,
+                }
+            },
+            default_method_media_type={
+                'PUT': 'application/json'
+            },
+            default_media_type='application/json',
+            **kwargs
+        )
+
```
nharraud: needs GET the one matching the provided ID, PATCH (modify the data with a JSON-patch)

b2share/modules/schemas/views.py
```diff
@@ -139,12 +147,96 @@ def get(self, community_schema):
...
+    def put(self, schema_id):
```
nharraud: PUT should modify the properties of the Block schema but not eh `BlockSchemaVersion`.
The current implementation should go in `BlockSchemaVersionResource`, check that the provided version doesn't already exist and follows the last one (last version +1)

b2share/modules/schemas/serializers.py
```diff
@@ -60,7 +70,24 @@ def block_schema_version_to_dict(block_schema_version):
...
+        block_schema_version
+            (:class:`b2share.modules.schemas.api:BlockSchemaVersion`):
+            block schema version that will be serialized.
+
+        code: http response status.
+
+        headers: additional http response headers.
+
+    Returns:
+        response: serialized response from BlockSchemaVersion.
```
nharraud: response is not a type, just the name of the retuned variable.

b2share/modules/schemas/serializers.py
```diff
@@ -60,7 +70,24 @@ def block_schema_version_to_dict(block_schema_version):
...
+    """
+    if block_schema_version is not None:
+        response = jsonify(block_schema_version_to_dict(block_schema_version))
+    else:
+        response = jsonify({})
```
nharraud: What is the scenario in which the `block_schema_version` is `None` and it is not an error?

b2share/modules/schemas/serializers.py
```diff
@@ -110,3 +137,89 @@ def community_schema_to_json_serializer(community_schema, code=200,
     if headers is not None:
         response.headers.extend(headers)
     return response
+
+
+def new_schema_to_json_serializer(schema, code=200, headers=None):
+    """Serializes schema to response.
```
nharraud: =>`Serialize schema to response.` (without "*s*"). Same for other docstrings.
```
The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".
```
https://www.python.org/dev/peps/pep-0257/#specification

b2share/modules/schemas/api.py
```diff
@@ -224,8 +272,14 @@ def create_version(self, json_schema):
             ).filter(
                 BlockSchemaVersionModel.block_schema == self.model.id
             ).one().last_version
+
             new_version = (last_version + 1 if last_version is not None
                            else 0)
+            if version_number and new_version < version_number:
+                raise InvalidSchemaVersionError()
```
nharraud: The exception should either be specific to "skipping version number" or it should receive a message explaining why the version number is invalid, i.e: "Provided version number should follow the last existing version number, which currently is <VERSION_NB>"

b2share/modules/schemas/views.py
```diff
@@ -139,12 +178,132 @@ def get(self, community_schema):
...
+        community_id = request.args['community_id']
+        schemas = BlockSchema.get_all_block_schemas(community_id=community_id)
+        return self.make_response(
+            schemas=schemas,
+            code=200)
+
+    def post(self):
+        """Create a new schema."""
+        if request.content_type != 'application/json':
+            abort(415)
```
nharraud: You can test it but I think `ContentNegotiatedMethodView` takes care of checking that the content-type is valid.

b2share/modules/schemas/views.py
```diff
@@ -139,12 +178,132 @@ def get(self, community_schema):
...
+
+class BlockSchemaResource(ContentNegotiatedMethodView):
+    view_name = 'block_schema_item'
+
+    def __init__(self, **kwargs):
+        """Constructor."""
+        super(BlockSchemaResource, self).__init__(
+            method_serializers={
+                'GET': {
+                    'application/json': block_schema_to_json_serializer,
```
nharraud: If all the serializers are the same you don't need to use `method_serializers `. Its purpose is to override the default serializer for some specific HTTP methods.

Thus you could just have:
```python
super(BlockSchemaResource, self).__init__(
    serializers={
        'application/json': block_schema_to_json_serializer
    },
    default_media_type='application/json',
    **kwargs
)
```

b2share/modules/schemas/api.py
```diff
@@ -194,7 +196,53 @@ def create_block_schema(cls, community_id, name, id_=None):
...
+                        setattr(self.model, key, value)
+                db.session.merge(self.model)
+        except sqlalchemy.exc.IntegrityError as e:
+            raise InvalidBlockSchemaError() from e
+        return self
+
+    def patch(self, patch):
+        """Update the schema's metadata with a json-patch.
+        Args:
+            patch (str): json-patch which can modify the following fields:
```
nharraud: The patch is not of type `str` but `dict`

b2share/modules/schemas/errors.py
```diff
@@ -82,6 +82,38 @@ class BlockSchemaVersionIsReleased(B2ShareSchemasError):
...
+                is {0}."""
+
+    def __init__(self, schema_number):
+        self.schema_number = schema_number
+        super(InvalidSchemaVersionError, self).__init__(
+            self.MESSAGE.format(schema_number)
+        )
+
+
+class SchemaVersionExistsError(B2ShareSchemasError):
```
nharraud: The two exception class names look a little strange. Exception `SchemaVersionExistsError` is raised when the version is invalid. Thus it is kind of strange that it does not inherit the existing exception `InvalidSchemaVersionError `.
What about having a generic exception `InvalidSchemaVersionError` and an additional `SchemaVersionSkippedError`? Both `SchemaVersionSkippedError` and `SchemaVersionExistsError ` would inherit `InvalidSchemaVersionError`.

b2share/modules/schemas/serializers.py
```diff
@@ -110,3 +134,86 @@ def community_schema_to_json_serializer(community_schema, code=200,
     if headers is not None:
         response.headers.extend(headers)
     return response
+
+
+def new_schema_to_json_serializer(schema, code=200, headers=None):
```
nharraud: Why is it "new" schema and not "block" schema. Is used only for newly created schemas?

nharraud: Or rather, I don't see any difference with `block_schema_to_json_serializer`. Why not reuse it?

b2share/modules/schemas/views.py
```diff
@@ -139,12 +178,125 @@ def get(self, community_schema):
...
+    view_name = 'block_schema_item'
+
+    def __init__(self, **kwargs):
+        """Constructor."""
+        super(BlockSchemaResource, self).__init__(
+            serializers={
+                'application/json': block_schema_to_json_serializer
+            },
+            default_method_media_type={
+                'PUT': 'application/json'
```
nharraud: There is no need to specify `'PUT': 'application/json'` as the default_media_type is already `application/json`

b2share/modules/schemas/views.py
```diff
@@ -139,12 +178,125 @@ def get(self, community_schema):
...
+            method_serializers={
+                'GET': {
+                    'application/json': schemas_list_to_json_serializer,
+                },
+                'POST': {
+                    'application/json': new_schema_to_json_serializer,
+                }
+            },
+            default_method_media_type={
+                'GET': 'application/json',
```
nharraud: I don't think that this is needed as the default_media_type is already `application/json`

b2share/modules/schemas/views.py
```diff
@@ -139,12 +178,125 @@ def get(self, community_schema):
...
+            try:
+                db.session.rollback()
+            except Exception as e2:
+                raise e2 from e1
+            abort(500)
+
+
 blueprint.add_url_rule(
-    '/schemas/<string:schema_id>/versions/<int:schema_version_nb>',
+    '/schemas/<string:schema_id>/versions/<string:schema_version_nb>',
```
nharraud: Why are version number strings?

tests/b2share_unit_tests/schemas/test_schemas_rest.py
```diff
@@ -68,3 +71,161 @@ def test_valid_get(app, test_communities):
...
+                url_for(
+                    'b2share_schemas.block_schema_list',
+                    community_id=community_id2
+                ),
+                headers=headers)
+            assert res.status_code == 200
+            response_data = json.loads(res.get_data(as_text=True))
+            assert isinstance(response_data['schemas'], list)
+            assert len(response_data['schemas']) == 2
+            assert response_data['schemas'][0]['name'] == block_schema_2.name
```
nharraud: I don't see any sorting in `BlockSchema. get_all_block_schemas`. This test might fail if the schemas are sent in another order. The simplest solution would be to have a default sorting in `BlockSchema. get_all_block_schemas` based on the date of creation.

b2share/modules/schemas/api.py
```diff
@@ -194,7 +196,53 @@ def create_block_schema(cls, community_id, name, id_=None):
...
+        Raises:
+            jsonpatch.JsonPatchConflict: the json patch conflicts on the
+                block schema.
+            jsonpatch.InvalidJsonPatch: the json patch is invalid.
+            b2share.modules.block_schemas.errors.InvalidBlockSchemaError: The
+                block schema patch failed because the resulting block schema is
+                not valid.
+        """
+        data = apply_patch({'name': self
+        .model.name}, patch, True)
```
nharraud: Could you please split the line after `.model.name}`. Splitting it on the dot makes the code less readable. It would also be better if the second line was indented properly.

b2share/modules/schemas/views.py
```diff
@@ -139,12 +183,128 @@ def get(self, community_schema):
         return community_schema
 
 
+class BlockSchemaListResource(ContentNegotiatedMethodView):
+    view_name = 'block_schema_list'
+
+    def __init__(self, **kwargs):
+        """Constructor."""
+        super(BlockSchemaListResource, self).__init__(
+            method_serializers={
```
nharraud: The serializers are the same so we can use ["serializers" as above](https://github.com/EUDAT-B2SHARE/b2share/pull/986/files#diff-48b91218db4176f52bd6e00d476fd5b1L79)

b2share/modules/schemas/views.py
```diff
@@ -139,12 +183,128 @@ def get(self, community_schema):
...
+            else:
+                block_schema = block_schema.patch(data)
+            db.session.commit()
+            return self.make_response(
+                schema=block_schema,
+                code=200
+            )
+        except (JsonPatchConflict):
+            abort(409)
+        except (JsonPatchException, InvalidJsonPatch):
```
nharraud: `InvalidJsonPatch` extends `JsonPatchException`. Thus there is no need to catch both of them.

PR #990, opened 2016-07-07
Oauth2server access tokens

emanueldima:


b2share/modules/users/serializers.py
```diff
@@ -0,0 +1,82 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EUDAT B2Share.
+# Copyright (C) 2016 University of Tuebingen, CERN.
+# Copyright (C) 2015 University of Tuebingen.
```
nharraud: This file is new. The copyright shouldn't mention 2015.

PR #1000, opened 2016-07-12
Readd lowest requirements install to dockerfile

emanueldima:


dockerize/Dockerfile
```diff
@@ -28,7 +28,10 @@ RUN npm install --unsafe-perm && \
     node_modules/webpack/bin/webpack.js -p
 
 WORKDIR /eudat/b2share
-RUN pip install -r requirements.txt
+RUN pip install requirements-builder
```
jirikuncar: you might want to join `RUN` commands to avoid intermediate image caches.

Issue #199, opened 2014-03-24
Upload files from other service


stranak:
Allow to log in to other services that can host files and pick files to upload from there.

Investigate if it can be done with:
- Google Drive
- Box.com
- Microsoft OneDrive
- Dropbox
- _B2DROP_
  - ~~Should B2DROP support Shibboleth and be in EduGAIN, this should work within Single-Sign-On, after #147.~~ B2DROP can't use Shibboleth.

emanueldima:
Added B2Drop integration in #1001

PR #1001, opened 2016-07-13
B2drop integration

emanueldima:
related to https://github.com/EUDAT-B2SHARE/b2share/issues/199

### DIFF
--- a/b2share/modules/remotes/__init__.py
+++ b/b2share/modules/remotes/__init__.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EUDAT B2Share.
+# Copyright (C) 2016 University of Tuebingen, CERN.
+#
+# B2Share is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# B2Share is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with B2Share; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""B2SHARE"""
+
+from __future__ import absolute_import, print_function
+
+from .b2drop import B2DropClient
+from .ext import B2ShareRemotes
+
+__all__ = ('B2DropClient', 'B2ShareRemotes')
--- a/b2share/modules/remotes/views.py
+++ b/b2share/modules/remotes/views.py
@@ -1,0 +1,193 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EUDAT B2Share.
+# Copyright (C) 2016 University of Tuebingen, CERN.
+#
+# B2Share is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# B2Share is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with B2Share; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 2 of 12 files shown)


b2share/modules/remotes/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""B2SHARE"""
+
+from __future__ import absolute_import, print_function
+
+from .b2drop import B2DropClient
+from .ext import B2ShareRemotes
+
+__all__ = ('B2DropClient', 'B2ShareRemotes')
```
nharraud: Why not add a `B2Drop` module? What is the purpose of the `B2ShareRemotes`? Will it contain other remote services?

b2share/modules/remotes/views.py
```diff
@@ -0,0 +1,193 @@
...
+            },
+            default_media_type='application/json',
+            **kwargs
+        )
+
+    def put(self, service):
+        if service == 'b2drop':
+            auth = request.get_json()
+            config = current_app.config.get('B2DROP_SERVER')
+            b2drop_client = B2DropClient(username=auth.get('username'),
```
nharraud: @thijscobben We would need an OAuth server on B2Drop to avoid this nasty hack.
Where can we create issues for B2Drop? I see many projects in the B2Drop github organization.

b2share/modules/remotes/__init__.py
emanueldima: Yes, we will eventually add dropbox/google cloud/box etc support. So it makes sense to have a common endpoint for this.

b2share/modules/remotes/views.py
```diff
@@ -0,0 +1,193 @@
...
+        b2drop_client = session.get('b2drop_client')
+        if not b2drop_client:
+            raise UserError("B2DROP remote not initialized")
+
+        stream = b2drop_client.make_stream_object(b2drop_path)
+        return put_file_into_bucket(bucket_id, key, stream, stream.length())
+
+
+def put_file_into_bucket(bucket_id, key, stream, content_length):
+    # TODO: refactor invenio_files_rest to have a proper API and use that one here
```
nharraud: Note: Is the API function still missing in the new invenio-files-rest? If not, is there an issue on invenio-files-rest?

b2share/modules/remotes/views.py
```diff
@@ -0,0 +1,193 @@
...
+            obj.set_contents(
+                stream, size=content_length, size_limit=size_limit)
+        db.session.commit()
+        return obj
+
+    return create_object(bucket, key)
+
+
+blueprint.add_url_rule('/remotes/jobs',
+                       view_func=Jobs.as_view(Jobs.view_name))
```
nharraud: There doesn't seem to be an endpoint to list currently running jobs. This would be useful for long running jobs or even the UI to see a progress bar.

b2share/modules/remotes/views.py
emanueldima: I didn't find a direct way in invenio-files-rest a4. I did not create an issue for this.

b2share/modules/remotes/views.py
emanueldima: The API is synchronous, for the moment. The `POST /jobs` call lasts until the transfer is over. This way a listing function is not needed, and we don't have to keep any state.

nharraud: Ok. This might be an issue if someone tries to upload a big file because it would stop the whole server. The scripts in `dockerize`run the server with `b2share run` and not a real webserver such as gunicorn or tornado. Thus it will block the whole testing server.
Can you add a file limit size so that the tests can run smoothly? We will remove it later.

nharraud:
Apart from the last comment I don't see any major issue.

b2share/modules/remotes/views.py
emanueldima: You're right, I should add a file size. But don't we have the exact same problem with any uploaded file?

nharraud: Yes I think you are right. In fact you don't have to add the max file size in your code. We can use [`FILES_REST_DEFAULT_MAX_FILE_SIZE ` config variable](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/config.py#L46) from invenio-files-rest. We just need to set it for the demo.

emanueldima: Ok, let's set it to.. 2GB? This is the limit supported by v1.

nharraud: At least for the testing server we should lower this number until we use a proper webserver. Imagine if B2Drop starts to upload slowly, then we might have a B2Share test server down for hours.

b2share/modules/remotes/__init__.py
emanueldima: It looks like uploading files from dropbox/owncloud will be eventually directly supported by invenio-files-rest: https://github.com/inveniosoftware/invenio-files-rest/issues/40

PR #1002, opened 2016-07-13
Add epic pid support

emanueldima:


emanueldima:
Note: b2share_epic.py has been copied with minimal modification from v1, and httplib2 has been added because of it

b2share/modules/records/b2share_epic.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def createHandle(location, checksum=None, suffix=''):
+    """ Create a new handle for a file.
+
+    Parameters:
+     location: The location (URL) of the file.
+     checksum: Optional parameter, store the checksum of the file as well.
+     suffix: The suffix of the handle. Default: ''.
+     Returns the URI of the new handle, raises a 503 exception if an error occurred.
```
nharraud: Could you please fix the comments here so that they follow the google comment styling (which we use with the napoleon sphinx plugin) http://google.github.io/styleguide/pyguide.html?showone=Comments#Comments
https://pypi.python.org/pypi/sphinxcontrib-napoleon

```
     Returns:
        the URI of the new handle, raises a 503 exception if an error occurred.
```

b2share/modules/records/minters.py
```diff
@@ -71,3 +74,29 @@ def b2share_oaiid_minter(record_uuid, data):
...
+
+    endpoint = 'b2share_records_rest.b2share_record_item'
+    url = url_for(endpoint, pid_value=local_pid.pid_value, _external=True)
+    url = url.replace('/api/records/', '/records/')
+    throw_on_failure = current_app.config.get('CFG_FAIL_ON_MISSING_PID', True)
+
+    try:
+        pid = createHandle(url)
+        if pid is None and throw_on_failure:
+            raise Exception("EPIC PID allocation failed")
```
nharraud: Note: we should have a common exception for PID registration failure.

b2share/modules/records/serializers/schemas/json.py
```diff
@@ -43,11 +43,15 @@ def filter_internal(self, data):
         if '_files' in data['metadata']:
             del data['metadata']['_files']
         if '_pid' in data['metadata']:
+            epic_pids = [p for p in data['metadata']['_pid']
+                         if p.get('type') == 'handle_pid']
+            if len(epic_pids) > 0:
+                data['metadata']['PID'] = epic_pids[0].get('value')
```
nharraud: It might be confusing for users to have things called just `PID`. Could we name it `handle_pid`? Or is it for backward compatibility?

nharraud:
Appart from the first comment and few questions it looks good.

b2share/config.py
```diff
@@ -190,3 +190,13 @@
...
     },
 }
+
+
+CFG_HANDLE_SYSTEM_BASEURL = 'http://hdl.handle.net'
+CFG_FAIL_ON_MISSING_PID = False
+
+# The following config variables are used to create EPIC PIDs
+# CFG_EPIC_USERNAME = 0000
+# CFG_EPIC_PASSWORD = ''
```
nharraud: are this account and password still valid? if so they should be removed. But also, why do we need to add this commented code?

b2share/modules/records/serializers/schemas/json.py
emanueldima: That's how it is in v1, but you might be right. Issue: https://github.com/EUDAT-B2SHARE/b2share/issues/1015

b2share/config.py
emanueldima: No, it's just a template, not a real configuration. These variables must be configured for the PIDs to be allocated.

b2share/modules/records/minters.py
emanueldima: I replaced it with the same EpicPIDException I defined earlier.

Issue #854, opened 2016-02-24
rest api: Uniform returned json structure for all search functions


emanueldima:
Better to follow invenio here, with this structure:

````
{
  "hits": {
    "total": 42, # total number of results, not just on this "page"
    "hits": [
      {
        #record, or community, or ..., as serialized for the GET
      }
    ]
  },
 "links": {
   "next": <URL>,
   "prev": <URL>
 }
}````

emanueldima:
closed by #1003

PR #1003, opened 2016-07-13
reshape returned communities json to conform with records and tokens

emanueldima:
closes https://github.com/EUDAT-B2SHARE/b2share/issues/854


nharraud:
@emanueldima we can merge this for now but this should be refactored later. The serializers should be used instead.
```python
method_serializers={
                 'GET': {
                     'application/json': search_to_json_serializer,
                 },
...
```

PR #1004, opened 2016-07-13
global: basic record & deposit permissions support

nharraud:
* NEW Adds record and deposit permission support.


nharraud:
Tests added. This PR adds basic permissions for records and deposit. I won't add the files permissions as I don't have enough time to finish it and prefer to not break everything before the RC1.

PR #1010, opened 2016-07-14
show correct version and git commit in web ui

emanueldima:


nharraud:
Ok for this change but then it might be good to handle versions the same way as we do in Invenio modules, ie bump `version.py` to a dev version just after the release.

You can for example take a look at the history of invenio-records: https://github.com/inveniosoftware/invenio-records-rest/commits/master
You will see
- [a commit named `release: v1.0.0a17`](https://github.com/inveniosoftware/invenio-records/commit/cb4b424d20afbb555a8e4c1d8c77223e0d9c0ab0) which sets `version.py` from `"1.0.0a17.dev20160623"` to `"1.0.0a17"`
- followed by [a commit named `version: post-release bump`](https://github.com/inveniosoftware/invenio-records/commit/aab3b024d0078ccd36334e1d2faae02d078c8882) which sets `version.py` from `"1.0.0a17"` to `"1.0.0a18.dev20160705"`

This enables to make the difference between released versions and development versions.

emanueldima:
ok: https://github.com/EUDAT-B2SHARE/b2share/issues/1014

PR #1011, opened 2016-07-14
Communities Command Line Interface

thijscobben:


thijscobben:
Now it needs (unit) tests.

emanueldima:
Please rebase 😄

b2share/modules/communities/api.py
```diff
@@ -91,12 +91,21 @@ def get(cls, id=None, name=None, with_deleted=False):
...
         """Searches for matching communities."""
-        start = int(start)
-        stop = int(stop)
         from .models import Community as CommunityMeta
-        metadata = CommunityMeta.query.order_by(CommunityMeta.created).limit(stop)[start:]
+        if (start==None and stop==None):
+            metadata = CommunityMeta.query.order_by(CommunityMeta.created)
+        elif (start and stop):
+            try:
+                start = int(start)
```
nharraud: The programmatic api shouldn't have to convert given parameters. It's the REST API and CLI job. The conversion can just be removed.

b2share/modules/communities/api.py
```diff
@@ -91,12 +91,21 @@ def get(cls, id=None, name=None, with_deleted=False):
...
-    def get_all(cls, start, stop):
+    def get_all(cls, start=None, stop=None):
         """Searches for matching communities."""
-        start = int(start)
-        stop = int(stop)
         from .models import Community as CommunityMeta
-        metadata = CommunityMeta.query.order_by(CommunityMeta.created).limit(stop)[start:]
+        if (start==None and stop==None):
+            metadata = CommunityMeta.query.order_by(CommunityMeta.created)
+        elif (start and stop):
```
nharraud: `bool(start) == False` does not mean that `start is None`
[It is recommended.](http://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html?highlight=none) to use `is None` or `is not None` in python.

nharraud:
Can we merge all the commits in this PR? It seems that some of them are just editing the same lines multiple times.

b2share/modules/schemas/root_schemas/bsv/bbmri.json
```diff
@@ -0,0 +1,89 @@
+{
```
nharraud: The root_schema directory should only contain "root schemas" released with B2Share. No community specific schema should be part of the default B2Share.
However specific schemas can be added to the demo just for demonstration purpose.

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+    click.secho("The following block schema versions are listed for community %s, community schema version %s" % (comm.name, "latest %d" % community_schema.version if not version else version))
+    for key in props:
+        click.secho("Block schema: %s, version url: %s" % (key, props[key]['$ref']))
+    
+@schemas.command()
+@with_appcontext
+@click.option('-v','--verbose', is_flag=True, default=False)
+@click.argument('community')
+@click.argument('block_schema_id')
+@click.option('--version')
```
nharraud: trailing white-spaces.

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+    block_schema.update(data)
+    db.session.commit()
+        
+@schemas.command()
+@with_appcontext
+@click.option('-v','--verbose', is_flag=True, default=False)
+@click.argument('block_schema_id')
+@click.argument('json_file')
+def block_schema_create_version(verbose, block_schema_id, json_file):
+    """Assign a json-schema file conforming to the EUDAT block-schema part of the root-schema that defines metadata format for this block schema. Upon success, this creates a new version, the old versions are maintained and can be obtained by block_schema_list_versions subcommand."""
```
nharraud: Can we keep the 80 column restriction? This will make back-porting code to invenio easier.
In invenio we follow [PEP-8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) and PEP-257.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EUDAT B2Share.
+# Copyright (C) 2016 University of Tuebingen, CERN.
+# Copyright (C) 2015 University of Tuebingen.
```
nharraud: This is a new file. There was no copyright in 2015. You can also add SURFsara in the list of organizations.

b2share/modules/communities/api.py
emanueldima: if start == 0 this fails, right? and it shouldn't.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+from os.path import isfile
+
+import click
+from flask_cli import with_appcontext
+
+from invenio_db import db
+
+from .api import Community, CommunityDoesNotExistError
+
+#from .errors import RootSchemaAlreadyExistsError
```
emanueldima: no commented out code in final PR, please

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+@click.argument('name')
+@click.argument('description')
+@click.argument('logo')
+def create(verbose, name, description, logo):
+    """Creating a community in the database. Name can be 255 characters long. Description is a text of maximally 1024 characters enclosed in parentheses. The logo parameter should be a valid path to a logo file. """
+    if len(name)>255:
+        raise click.BadParameter("NAME parameter is longer than the 255 character maximum")
+    if len(description)>1024:
+        raise click.BadParameter("DESCRIPTION parameter is longer than the 1024 character maximum")
+    if not isfile(logo):
```
emanueldima: The logo is used as image source in the UI, so it MUST be a correct URL. The `isfile(logo)` call does not take into account the current working directory. The test should be something like:

````
isfile(os.path.join(os.environ.get('B2SHARE_UI_PATH'), logo))
````

File: b2share/modules/communities/cli.py [hunk 9]
```diff
@@ -0,0 +1,120 @@
...
+        if community_exists:
+            raise click.BadParameter("There already is a community with name %s" % name)
+    except CommunityDoesNotExistError as e:
+        pass
+    try:
+        community = Community.create_community(name=name,description=description,logo=logo)
+        db.session.commit()
+        if verbose:
+            click.echo("Community created with %d" % community.id)
+    except Exception as e1:
```
emanueldima: Is this useful? If the creation fails, you'd get an exception and the rollback is automatic.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+        raise click.ClickException(str(e1))
+        
+@communities.command()
+@with_appcontext
+@click.option('-v','--verbose', is_flag=True, default=False)
+def list(verbose):
+    """List all communities in this instances' database"""
+    communities = Community.get_all()
+    for c in communities:
+        click.echo("%s\t%s\t%s\t%s" % (c.name, c.id, c.description, c.logo))
```
emanueldima: Tabs here are not very useful, as the name and description should be quite long. I'd use newlines.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+@click.argument('id')
+def edit(verbose, id, name, description, logo,clear_fields):
+        """Edit data of the specified community."""
+        try:
+            community = Community.get(id=id)
+        except:
+            #CommunityDoesNotExistError, ValueError, sqlalchemy.exc.StatementError
+            raise click.BadParameter("No community with id %s" % id)
+        if not(name or description or logo):
+            raise click.ClickException("At least one of name, description or id must be specified")
```
emanueldima: 'name, description or logo' instead of 'name, description or id'

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+        data = {}
+        if name:
+            if not(name==community.name):
```
emanueldima: why not `if name != community.name`?

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+                try:
+                    Community.get(name=name)
```
emanueldima: This line will get the community named by 'name', if any. It will not raise an error if the community exists, on the contrary, it will raise a CommunityDoesNotExistError.

b2share/modules/communities/errors.py
```diff
@@ -39,3 +39,7 @@ class CommunityDoesNotExistError(Exception):
 class CommunityDeletedError(Exception):
     """Exception raised when a requested community is marked as deleted."""
     pass
+
+class GetAllStartStopError(Exception):
```
emanueldima: I'd say this error is too specific, how about one named InvalidArgumentsError? A more general, reusable error may be more useful.

b2share/modules/communities/helpers.py
```diff
@@ -0,0 +1,37 @@
...
+from .api import Community
+from .errors import CommunityDoesNotExistError
+
+def get_community_by_name_or_id(community):
+    result = None
+    try:
+        result = Community.get(id=community)
+    except:
+        try:
+            #TODO: case insensitivity?
```
emanueldima: The case sensitivity should be handled by Community.get

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+        click.secho("%s\t%s\t%s\t%s\t\t%d" % (bs.id,bs.name[0:15].ljust(15), bs_comm.name[0:15].ljust(15), bs.deprecated, len(bs.versions)  ))
+        
+@schemas.command()
+@with_appcontext
+@click.option('-v','--verbose', is_flag=True, default=False)
+@click.option('-n','--name',help='set the name of the community')
+@click.option('-c','--community',help='set the maintaining community by name or id')
+@click.option('-d','--deprecated',help='(un)set deprecated bit, 1 is deprecated, 0 is not deprecated')
+@click.argument('block_schema_id')
+def block_schema_edit(verbose, name, community, deprecated, block_schema_id):
```
emanueldima: I'd put block_schema_id after verbose, not at the end of the list

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+    db.session.commit()
+        
+@schemas.command()
+@with_appcontext
+@click.option('-v','--verbose', is_flag=True, default=False)
+@click.argument('block_schema_id')
+@click.argument('json_file')
+def block_schema_create_version(verbose, block_schema_id, json_file):
+    """Assign a json-schema file conforming to the EUDAT block-schema part of the root-schema that defines metadata format for this block schema. Upon success, this creates a new version, the old versions are maintained and can be obtained by block_schema_list_versions subcommand."""
+    try:
```
emanueldima: this try could be abstracted into a `check_uuid` function

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+        val = UUID(block_schema_id, version=4)
+    except ValueError:
+        raise click.BadParameter("BLOCK_SCHEMA_ID is not a valid UUID (hexadecimal numbers and dashes e.g. fa52bec3-a847-4602-8af5-b8d41a5215bc )")
+    try:
```
emanueldima: same here, this could be a function called `checked_get_block_schema` or similar

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+    #save
+    CommunitySchema.create_version(comm.id, community_schema_dict)
+    db.session.commit()
+    community_schema = CommunitySchema.get_community_schema(comm.id)
+    if verbose:
+        click.secho("CommunitySchema created for %s , version %d" % (comm.name, community_schema.version))
+        
+def community_schema_remove_block_schema_version(verbose, block_schema_id, version=None):
+    """Creates a new version of the community schema with the specified (version of the) block schema removed."""
+
```
emanueldima: #TODO ?

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+    version_url = resolve_schemas_ref("$BLOCK_SCHEMA_VERSION_URL[%s::%d]" % (block_schema_id, version))
+    #fetch community_schema
+    community_schema = CommunitySchema.get_community_schema(comm.id)
+    #No community schema? Create one.
+    if not community_schema:
+        version_dict = {}
+        ref_dict = {}
+        ref_dict['$ref'] = version_url
+        version_dict[block_schema_id] = ref_dict
+        props_dict[block_schema_id] = version_dict
```
emanueldima: What's this `props_dict`, it's not defined anywhere else, is it?

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+        community_schema_dict = {}
+        community_schema_dict['properties'] = props_dict
+        community_schema_dict['$schema'] = "http://json-schema.org/draft-04/schema#"
+        community_schema_dict['type'] = "object"
+        community_schema_dict['additionalProperties'] = "false"
```
emanueldima: Why not rewrite the entire block as a single dict, something like:
````
        community_schema_dict = {
               "$schema": "http://json-schema.org/draft-04/schema#",
                "type": "object",
                "properties": {
                    block_schema_id: {
                        "$ref": version_url,
                    }
                },
                "additionalProperties": false
````
I think it would be much more clear this way.

File: b2share/modules/communities/api.py [hunk 22]
```diff
@@ -91,12 +91,21 @@ def get(cls, id=None, name=None, with_deleted=False):
...
+            try:
+                start = int(start)
+                stop = int(stop)
+            except ValueError as e:
+                raise GetAllStartStopError("start and stop parameters must be numbers")
+            metadata = CommunityMeta.query.order_by(CommunityMeta.created).limit(stop)[start:]
+        else:
+            #one of them is None this cannot happen
+            raise GetAllStartStopError("Neither or both start and stop should be None")
```
nharraud: There is a standard error type for "invalid argument value": [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@communities.command()
+@with_appcontext
+@click.option('-v', '--verbose', is_flag=True, default=False)
+@click.argument('name')
+@click.argument('description')
+@click.argument('logo')
+def create(verbose, name, description, logo):
+    """Creating a community in the database. Name can be 255 characters long. Description is a text of maximally 1024 characters enclosed in parentheses. The logo parameter should be a valid path to a logo file. """
```
nharraud: s/Creating/Create/g

In PEP-257: *The docstring is a phrase ending in a period. [It prescribes the function or method's effect as a command](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings) ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".*

Regarding the additional information, such as parameter types and description, [we use sphinx napoleon plugin](http://www.sphinx-doc.org/en/stable/ext/napoleon.html) which enables us to use [google style format for docstrings](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments)

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+    if len(name)>255:
+        raise click.BadParameter("NAME parameter is longer than the 255 character maximum")
+    if len(description)>1024:
+        raise click.BadParameter("DESCRIPTION parameter is longer than the 1024 character maximum")
+    if not isfile(logo):
+        raise click.BadParameter("LOGO should be a (relative) path to an existing image file.")
+    try:
+        community_exists = Community.get(name=name)
+        if community_exists:
```
nharraud: `CommunityDoesNotExistError ` will always be raised when the community is not found. Thus there is no need for this test.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+            raise click.BadParameter("There already is a community with name %s" % name)
+    except CommunityDoesNotExistError as e:
+        pass
+    try:
+        community = Community.create_community(name=name,description=description,logo=logo)
```
nharraud: space after comma for consistency.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+def list(verbose):
+    """List all communities in this instances' database"""
+    communities = Community.get_all()
+    for c in communities:
+        click.echo("%s\t%s\t%s\t%s" % (c.name, c.id, c.description, c.logo))
+
+
+@communities.command()
+@with_appcontext
+@click.option('-v','--verbose',is_flag=True,default=False)
```
nharraud: space after comma for consistency.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+@click.option('--name')
+@click.option('--description')
+@click.option('--logo')
+@click.option('--clear_fields',is_flag=True,default=False,help='if set edit nullifies unspecified value options')
```
nharraud: space after comma for consistency.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,120 @@
...
+        if not(name or description or logo):
+            raise click.ClickException("At least one of name, description or id must be specified")
+        data = {}
+        if name:
+            if not(name==community.name):
+                try:
+                    Community.get(name=name)
+                except CommunityExistsError:
+                    raise click.BadParameterException("You are trying to change the name of the community but another community already exists with that name.")            
+            data['name']=name
```
nharraud: PEP8: ["*Always surround these binary operators with a single space on either side: assignment ( = )...*"](https://www.python.org/dev/peps/pep-0008/#other-recommendations)

b2share/modules/communities/errors.py
nharraud: As said above, [The generic python `ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) does exactly that.

b2share/modules/communities/helpers.py
nharraud: I wouldn't try to make this too complex for now. This CLI will **only** be used by the admin. Most of the time he will probably use the ID.

Also the current `name` of the community will probably be matched to the `title` field in invenio-communities model. This field has a length of 255 characters. I doubt that the admin will use such a long value. Maybe later we can see if a prefix can be used when it is unique, but this is more a bonus feature if it is asked for.

b2share/modules/schemas/cli.py
```diff
@@ -27,10 +27,28 @@
...
+from .helpers import load_root_schemas, resolve_schemas_ref
+from .validate import validate_metadata_schema
+from .api import BlockSchema, CommunitySchema
+
+from b2share.modules.communities.api import Community, CommunityDoesNotExistError
+from b2share.modules.communities.cli import communities
+from b2share.modules.communities.helpers import get_community_by_name_or_id
+
+from b2share.modules.communities.api import Community, CommunityDoesNotExistError
+from b2share.modules.communities.cli import communities
```
nharraud: duplicate imports

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+    for bs in block_schemas:
+        bs_comm = Community.get(id=bs.community)
+        click.secho("%s\t%s\t%s\t%s\t\t%d" % (bs.id,bs.name[0:15].ljust(15), bs_comm.name[0:15].ljust(15), bs.deprecated, len(bs.versions)  ))
+        
+@schemas.command()
+@with_appcontext
+@click.option('-v','--verbose', is_flag=True, default=False)
+@click.option('-n','--name',help='set the name of the community')
+@click.option('-c','--community',help='set the maintaining community by name or id')
+@click.option('-d','--deprecated',help='(un)set deprecated bit, 1 is deprecated, 0 is not deprecated')
```
nharraud: Rather than `bit` which is an implementation detail. We could just use the term `flag` which people tend to understand better.

The documentation would then be: "(un)mark the schema as deprecated. Values can be either `True` to add the deprecation flag, or `False` to remove the flag."

b2share/modules/schemas/cli.py
```diff
@@ -46,3 +64,225 @@ def init(verbose):
...
+    for key in props:
+        click.secho("Block schema: %s, version url: %s" % (key, props[key]['$ref']))
+    
+@schemas.command()
+@with_appcontext
+@click.option('-v','--verbose', is_flag=True, default=False)
+@click.argument('community')
+@click.argument('block_schema_id')
+@click.option('--version')    
+def community_schema_add_block_schema_version(verbose, community, block_schema_id, version=None):
```
nharraud: I am not sure about `community_schema_remove_block_schema_version` and `community_schema_add_block_schema_version`. It adds features which
1/ are not supported by the programmatic and REST APIs.
2/ we don't know if the admin will have a need for. The CLI is **only** for the admin/superuser having direct access to the b2share server.

Also the CLI does not have the simpler community schema edition functions provided by the programmatic and REST APIs, i.e. just giving json directly with the root schema version number. Thus it is not possible to later use "AllOf", "AnyOf", "OneOf" or more complex structures. It is also not possible to update the root schema version used by a community schema.

I understand the need for simplicity, and I agree that there is a use case for "adding" and "removing" block schemas to community schemas. I am just not sure that this is the right moment and place to add this feature. There is a risk of constraining what we can do in the future by implementing it now like that.

A simpler step, in my humble and subjective opinion, would be to do the first implementation of this advanced feature in the UI. We can suppose that the admin knows a little about JSON and will be able to write the JSON Schema if he needs to.  That way the REST API and the CLI remain simple and we can migrate the feature in the backend later once we have more user feedback, and if it is needed.

b2share/modules/communities/api.py
thijscobben: OK I removed the check. CLI doesn't use these parameters anyway. However, communites/views/CommunityListResource/get calls with request.args.get start and stop parameters that I think are not type checked line 148 of api.py .

nharraud: @thijscobben Indeed. This REST API code should also be refactored in order to follow [the same pagination style as in invenio-records-rest](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L429), i.e. `page` and `size`.

b2share/modules/communities/cli.py
thijscobben: OK I check whether the file exists in
B2SHARE_UI_PATH/img/communities (as the existing community logo files are there)
Can anyone advise me if this directory is persistent over re-installs ? (when UI install copies files?)

emanueldima: This path should always exist,  with the images. They are cloned directly from github.

nharraud: @emanueldima What do you mean by *cloned from gihub*? We cannot know in advance the logo of the communities.

emanueldima: I assumed someone would place them in the right folder, since we're talking about administrators doing this operation. An alternative would be to accept a full URL, download the image stored at that url, and place it in the right folder. This would be certainly a nice to have feature.

nharraud: Ah ok. We could for now keep things simple and suppose that the admin copied the file in the right directory.
invenio-communities already handles some [complex logo use cases](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/models.py#L373) by using a file bucket. The less we do the less we will have to refactor.

b2share/modules/communities/cli.py
thijscobben: Added some formatting. I like it tabular.

b2share/modules/communities/cli.py
thijscobben: This is for duplicate name checking.

b2share/modules/schemas/cli.py
thijscobben: We talked about this before. Something is to be said for your position. I propose to leave this code 'as-is' and to raise the discussion later with users how they would like to manage 'community_schema" element in their metadata schemas.

b2share/modules/communities/cli.py
```diff
@@ -0,0 +1,127 @@
...
+           raise click.BadParameter("No community with id %s" % id)
+        if not(name or description or logo):
+            raise click.ClickException("""At least one of name, description or 
+            id must be specified""")
+        data = {}
+        if name:
+            if not(name==community.name):
+                try:
+                    Community.get(name=name)
+                except CommunityExistsError:
```
nharraud: I can't find where this exception is defined. Shouldn't it be `CommunityDoesNotExistError` ?

b2share/modules/schemas/cli.py
thijscobben: I tried to make the code more readable by implementing dictionaries of dictionaries on separate lines and prevent code like
community_schema_dict['properties'][str(block_schema_id)]={str(block_schema_id):{str(block_schema_id):{'$ref':version_url}}}

b2share/modules/schemas/cli.py
thijscobben: I removed the code and I added an issue #1098 .

PR #1027, opened 2016-07-19
global: temporary serialization fix

nharraud:
* Temporary fix for the serialization exception happening when
  a record GET follows an /api/records/ search.


emanueldima:
Ok, I tested it on my dockerize environment and it works correctly.

PR #1036, opened 2016-07-22
Adding select language component

SarahBA:


b2share/factory.py
```diff
@@ -116,3 +116,7 @@ def serve_records(path):
     @app_ui.route('/search/<path:path>')
     def serve_search(path):
         return app_ui.send_static_file('index.html')
+
+    @app_ui.route('/languages.json', defaults={'path': ''})
```
emanueldima: I would add here a /static endpoint for serving static files, and languages.json would be only one of them (disciplines would be another). like this:
````
@app_ui.route('/static/<path:filename>')
    def serve_static_resources(filename):
        return app_ui.send_from_directory('static', filename)
````

File: devenv/run_demo.sh [hunk 2]
```diff
@@ -83,8 +83,8 @@ fi
 cdvirtualenv src/b2share/devenv
 if [ -n "$REINIT" ]; then
 	echo; echo "### Reinitialize elasticsearch and redis"
-	docker-compose down
-	docker-compose down
+	docker-compose stop
```
emanueldima: Not stop, no. We want down so that the elasticsearch and redis containers will be reinitialized completely.

webui/package.json
```diff
@@ -14,6 +14,7 @@
     "babel-preset-es2015": "^6.1.18",
     "babel-preset-react": "^6.1.18",
     "css-loader": "^0.23.0",
+    "json-loader": "^0.5.4",
```
emanueldima: This is not needed if the languages are queried on demand.

webui/webpack.config.devel.js
```diff
@@ -9,12 +9,18 @@ module.exports = {
...
         new WebpackShellPlugin({onBuildStart:['./make_version_js.sh']})
     ],
+    resolve: { extensions: ['', '.js', '.jsx', '.json'] },    
     module: {
         loaders: [
             {   test: /\.jsx?$/,
                 loader: 'babel-loader',
                 query: { presets: ['es2015', 'react'] },
                 include: path.join(__dirname, 'src')
+            },
```
emanueldima: This should not be needed

webui/src/data/server.js
```diff
@@ -510,14 +510,17 @@ class ServerCache {
             ajaxGet({
                 url: apiUrls.languages(),
                 successFn: (data) => {
-                    const langs = data.languages.map(([id, name]) => ({id, name}));
+                    const langs = data.languages.map(([id,name]) => ({"id": id, "name": name}));
```
emanueldima: The original is perfectly valid, see here: https://github.com/lukehoban/es6features#enhanced-object-literals

webui/src/data/server.js
```diff
@@ -510,14 +510,17 @@ class ServerCache {
...
                     console.log('got languages: ', langs.length);
                     this.store.setIn(['languages'], langs);
                 },
-                errorFn: (xhr) => this.store.setIn(['languages'], new Error(xhr)),
+                errorFn: (xhr) => {
+                    console.log("Something wrong: " + xhr)
```
emanueldima: Don't show it in the final PR 😄

webui/src/components/selectlanguage.jsx
```diff
@@ -0,0 +1,76 @@
...
+            && this.state.langNum < this.state.maxLang)
+        {
+            this.state.langNum = this.state.langNum+1;
+            ret = true
+        }
+        return ret;
+    },
+
+    render() {
+        const langgg = serverCache.getLanguages();
```
emanueldima: naming! 'languages' is a bit longer, but is more readable -- or you could use langs

webui/src/components/selectlanguage.jsx
```diff
@@ -0,0 +1,76 @@
...
+        if (!langgg) {
+            return <Wait/>;
+        }
+        else{
+            var def = this.props.defValue;      
+            if (!def){
```
emanueldima: this needs to be simplified: 

````
var def = this.props.devValue || 'eng';
````

File: webui/src/components/selectlanguage.jsx [hunk 9]
```diff
@@ -0,0 +1,76 @@
...
+                var defaultVal = langgg[1833].id;  
+            }
+            else{
+                const loc = langgg.filter(i => i.id === def);
+                var defaultVal = loc[0].id;
+            }
+
+            this.state.value = defaultVal;
```
emanueldima: you shouldn't assign to state, use setState instead

PR #1061, opened 2016-07-29
api: user identity updated by access token fix

nharraud:
* FIX Uses the last fix in invenio-oauth2server which sets the
  user identity when the API is accessed with an access token.


nharraud:
The PR is ready but I am waiting for inveniosoftware/invenio-oauth2server#65 to be merged.

PR #1072, opened 2016-08-09
Adding select language component

SarahBA:


webui/src/components/selectlanguage.jsx
```diff
@@ -0,0 +1,72 @@
...
+        return ret;
+    },
+
+    componentWillMount() {
+        this.componentWillReceiveProps(this.props);
+    },
+
+    componentWillReceiveProps(props) {
+        var def = this.props.defValue || 'eng'; 
+        this.setState({value:def});
```
SarahBA: @emanueldima There is a problem here that I couldn't fix. When the user choose a new value, the dropdown menu doesn't show it, while the parent is getting the new value. Can you please take a look?

webui/src/components/editrecord.jsx
```diff
@@ -373,6 +379,22 @@ const EditRecord = React.createClass({
...
+        if (fieldID == "language_code") {
+             return (
+                <div className="form-group row" key={fieldID} style={{marginTop:'1em'}} title={fieldSchema.get('description')}>
+                    <label htmlFor={fieldID} className="col-sm-3 control-label" style={{fontWeight:'bold'}}>
+                        <span style={{float:'right', color:isError?'red':'black'}}>
+                            {fieldSchema.get('title') || fieldID} {type.required ? "*":""}
+                        </span>
+                    </label>
+                    <div className="col-sm-9" >
+                        <SelectLanguage onSelectLanguage={this.handleLanguageCode} defValue={getValue()} />
```
emanueldima: Try using `onSelectLanguage={setValue}`. setValue is defined above and sets the language code directly in the record.

webui/src/components/selectlanguage.jsx
emanueldima: It's probably related to the setvalue comment, see above.

webui/src/components/editrecord.jsx
```diff
@@ -183,6 +184,7 @@ const EditRecord = React.createClass({
         return {
             record: null,
             fileState: 'done',
+            language: '',
```
emanueldima: This is not needed. The language is set directly in the state.record.

webui/src/components/editrecord.jsx
```diff
@@ -340,6 +342,10 @@ const EditRecord = React.createClass({
         );
     },
 
+    handleLanguageCode: function(langValue) {
```
emanueldima: Not needed, see above.

webui/src/data/server.js
```diff
@@ -522,10 +522,13 @@ class ServerCache {
...
                 },
-                errorFn: (xhr) => this.store.setIn(['languages'], new Error(xhr)),
+                errorFn: (xhr) => {
+                    console.log("Couldn't get the languages: " + xhr)
+                    this.store.setIn(['languages'], new Error(xhr));
+                },                
             });
         }
-        return langs;
+        return this.store.getIn(['languages']);
```
emanueldima: why the change? it's not needed, and not optimal.

webui/src/data/server.js
```diff
@@ -522,10 +522,13 @@ class ServerCache {
                     console.log('got languages: ', langs.length);
                     this.store.setIn(['languages'], langs);
                 },
-                errorFn: (xhr) => this.store.setIn(['languages'], new Error(xhr)),
+                errorFn: (xhr) => {
+                    console.log("Couldn't get the languages: " + xhr)
```
emanueldima: try not committing any console.log calls 😄

webui/src/components/selectlanguage.jsx
SarahBA: The issue was due to using `componentWillReceiveProps`. It doesn't trigger a render after the change!

PR #1097, opened 2016-08-26
WIP Embargo + docker improvements

### LINKED ISSUE #1066: deployment: Use local clone of B2Share when building docker image
**Problem:**
Current **Dockerfile** creates the image using a clean clone of B2Share from Github. This prevents us from testing code before sending PRs.

Thus the Dockerfile should use the local clone of B2Share.

### LINKED ISSUE #1069: config: use JSONSCHEMAS_HOST instead of SERVER_NAME
**Why:**
- We have had many issues with `SERVER_NAME` because of the routing from nginx.
- `invenio-jsonschemas` uses `JSONSCHEMAS_HOST` to get the host of JSON Schemas. We should do the same so that we are more compatible with invenio.

nharraud:
* Adds a new draft_json_schema field in the communities' schema
  REST API result. This enables draft and published records
  schema separation.

* NEW Adds embargo date update on the backend. Front-end still doesn't
  support embargo properly.

* Uses local clone to build the docker image instead of fetching
  a new copy from Github. This enables better testing.
  (closes #1066)

* Removes references to SERVER_NAME and uses JSONSCHEMAS_HOST
  instead. (closes #1069)

* Improves docker images build by removing unnecessary scripts.
  Updates deployment documentation.


nharraud:
@llehtine Could you take a look at the last commit "global: docker deployment improvement"?
It is improving the docker deployment now that `SERVER_NAME` is not needed. I would like to know if this will affect the test server deployment.

llehtine:
Took a look on that one. Looks good, I would think the nginx server deployment is easier now with the config-file variables in use. I don't think it will break the test-server configuration

nharraud:
@llehtine Thanks for the review. I am now trying to replace redis with RabbitMQ for the celery tasks as the Kombu library fails each time on travis when using redis. I don't like to add a new process, especially so close to a release, but this one was either way planned to arrive sooner or later.

PR #1100, opened 2016-08-30
release: 2.0.0.rc4

nharraud:

### DIFF
--- a/b2share/version.py
+++ b/b2share/version.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of EUDAT B2Share.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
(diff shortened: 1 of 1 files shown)


b2share/version.py
```diff
@@ -20,4 +20,4 @@
 """Version number."""
 
 
-__version__ = "2.0.0.rc3"  # FIXME: generate the version number
+__version__ = "2.0.0.rc4"  # FIXME: generate the version number
```
jirikuncar: @nharraud it should be probably `2.0.0rc4` (see https://www.python.org/dev/peps/pep-0440/#pre-releases).

nharraud: @jirikuncar Thanks. I agree that we should follow python standard for release numbering. I'll check with @emanueldima what is the reason for this custom numbering when he is back, which means that will keep this version for this RC.

PR #1105, opened 2016-08-31
create record ends up with 400 BAD REQUEST

BeneDicere:
use open_access instead of open in json body


emanueldima:
rebased and merged manually, see https://github.com/EUDAT-B2SHARE/b2share/commit/efdbc2688f3db619187bb512e640a0c67f14276d

PR #1167, opened 2016-10-21
deploy: pinning of all python dependencies

### LINKED ISSUE #1165: deploy: pin dependencies version for the docker container
**problem:**
We are not pinning python dependencies recursively. We only use the lowest requirements when building the docker container. This is not enough as it can break when the libraries we depend on require different versions of the same library (see #1164).

**solution:**
Use `pip-compile` to pin the dependencies for the docker container.

**warning:**
We need to test the pinned requirements on travis.

nharraud:
* Pins all python dependencies. (closes #1165)

* Removes "lowest" tests from travis and replaces them with "release"
  tests which use pinned dependencies. Previous "release" tests,
  which were using latest version of dependencies, are renamed
  "current".


nharraud:
Onve this is merged I will update the ("Developer's corner" wiki page](https://github.com/EUDAT-B2SHARE/b2share/wiki/Developer's-corner) to describe how to update the pinned dependencies. This will be useful before releasing.

Pinning the dependencies also means that the docker container can break when B2Share code uses new modules. The container should be checked before each release.

dockerize/Dockerfile
```diff
@@ -1,14 +1,14 @@
...
+RUN rpm -iUvh http://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-8.noarch.rpm
+RUN yum -y update
+RUN yum -y install wget gcc-c++ openssl-devel postgresql-devel python34 python34-devel npm supervisor git libffi-devel python-devel libxml2-devel libxml2 libxslt-devel zlib1g-devel libxslt
 
 WORKDIR /tmp
-RUN wget https://bootstrap.pypa.io/get-pip.py && \
-    python3.4 get-pip.py && \
-    pip3 install --upgrade pip
+RUN wget https://bootstrap.pypa.io/get-pip.py
+RUN python3.4 get-pip.py
```
jirikuncar: @nharraud there is no need in creating new layers instead keep the `&&` pattern.

emanueldima:
I tested this PR locally, works fine.

PR #1169, opened 2016-10-24
http api documentation update

emanueldima:


webui/src/docs/rest-api.jsx
```diff
@@ -1,244 +1,540 @@
...
+        <p>To get an access token, login to the B2SHARE web interface and click
+            on your username. On the User Profile page, go to the “API Tokens”
+            section, enter a token identification name (e.g. 'api') and click
+            “New Token”. This will create an access token, visible on the
+            screen. Please note that this is the only time the access token is
+            visible, so copy it to a safe place. </p>
+        <p>The following shell commands will expect that the ACCESS_TOKEN
+            environment variable is defined and contains the actual
+            access_token. The command to define this variable is: </p>
+        <Example>export ACCESS_TOKEN='...'</Example>
```
nharraud: I would suggest to put a real access-token (not valid of course). The users will most probably know how to export env variables as they want to use the REST API. Thus this example would be more useful if it also showed what an access token looks like.

webui/src/docs/rest-api.jsx
```diff
@@ -1,244 +1,540 @@
...
+            <li><p>Required parameters: access_token</p></li>
+            <li><p>Returns: the JSON schema, embedded in a JSON object, or an error message.</p></li>
+        </ul>
+        <Example>
+            curl https://$HOSTNAME/api/communities/$COMMUNITY_ID/schemas/last?access_token=$ACCESS_TOKEN
+            <Returns>
+            {{
+              "community": "e9b9792e-79fb-4b07-b6b4-b9c2bd06d095",
+              "draft_json_schema": {
+                "$ref": "https://vm0045.kaj.pouta.csc.fi/api/communities/e9b9792e-79fb-4b07-b6b4-b9c2bd06d095/schemas/0#/json_schema",
```
nharraud: What about creating a fake simple hostname which we keep consistent through all the examples.
```sh
 <Example>export HOSTNAME='trng-b2share.eudat.eu'</Example>
...
"$ref": "https://trng-b2share.eudat.eu/api/communities/e9b9792e-79fb-4b07-b6b4-b9c2bd06d095/schemas/0#/json_schema",
```
It would IMHO be more clear.

webui/src/docs/rest-api.jsx
```diff
@@ -1,244 +1,540 @@
...
+                "self": "https://vm0045.kaj.pouta.csc.fi/api/records/?sort=bestmatch&q=community%3Ae9b9792e-79fb-4b07-b6b4-b9c2bd06d095&size=10&page=1"
+              }
+            }}
+            </Returns>
+        </Example>
+
+        <h3>Search records</h3>
+        <p>Search all the published records for a query string.</p>
+        <ul>
+            <li><p>URL path: /api/records/?q=$QUERY_STRING</p></li>
```
nharraud: This one is hard without knowing the search query syntax.

webui/src/docs/rest-api.jsx
```diff
@@ -1,244 +1,540 @@
...
+            <li><p>Returns: the list of matching records (in JSON format) or an error message.</p></li>
+        </ul>
+        <Example>
+            curl https://$HOSTNAME/api/records/?q=$QUERY_STRING?access_token=$ACCESS_TOKEN
+        </Example>
+
+        <h3>Search drafts</h3>
+        <p>Search for all drafts (unpublished records) that are accessible by the requestor. Usually this means own records only.</p>
+        <ul>
+            <li><p>URL path: /api/records/?drafts=1</p></li>
```
nharraud: normally there is no need to add the `=1` in `?drafts=1`. If it doesn't work without it it's a bug.

webui/src/docs/rest-api.jsx
```diff
@@ -1,244 +1,540 @@
...
+        <p>Create a new record, in the draft state.</p>
+        <ul>
+            <li><p>URL path: /api/records</p></li>
+            <li><p>HTTP method: POST</p></li>
+            <li><p>Required URL parameter: access_token</p></li>
+            <li><p>Required data payload: json object with basic information about the object </p></li>
+            <li><p>Returns: the new draft record contents and location. Please note that
+                the returned json object contains also the URL of the file bucket used for the record.
+                Also note that the URL of the draft record, needed for setting record metadata,
+                will end in '/drafts/'</p></li>
```
nharraud: `/draft/` not `/drafts/` (without `s`)

webui/src/docs/rest-api.jsx
```diff
@@ -1,244 +1,540 @@
...
+        </ul>
+        <Example>
+            {'curl -i -H "Content-Type:application/json" -d \'{"title":"TestRest", "community":"e9b9792e-79fb-4b07-b6b4-b9c2bd06d095", "open_access":true}\' -X POST https://$HOSTNAME/api/records/?access_token=$ACCESS_TOKEN'}
```
nharraud: The starting `{` in `{'curl -i -H ...` seems to be there by mistake.

webui/src/docs/rest-api.jsx
```diff
@@ -1,244 +1,540 @@
...
+            files and associated metadata. The record’s <strong>metadata </strong>
+            consists of a set of common fixed metadata fields and a set of custom
+            metadata blocks, each block containing related metadata fields. A
+            record is always connected to one scientific community which has the
+            role of curating and maintaining it.</p>
+        <p> A data record can exist in several states. Immediately after creation a record
+            enters the 'draft' state. In this state the record is only accessible
+            by its owner and can be freely modified: its metadata can be changed
+            and files can be uploaded into or removed from it. A draft can be
+            published at any time, and through this action it changes its state
```
emanueldima: Yes, we should document that, but after we have the feature. This doc should be in sync with the actual state of the product.

webui/src/docs/rest-api.jsx
emanueldima: Ok.

webui/src/docs/rest-api.jsx
emanueldima: Good find.

webui/src/docs/rest-api.jsx
emanueldima: No, this is valid, just part of the jsx special syntax. There is a corresponding bracket at the end.

PR #1171, opened 2016-10-25
Fix to saving large files from importer

thijscobben:


demo/b2share_demo/helpers.py
```diff
@@ -418,8 +419,12 @@ def _save_file(url, filename):
...
     while not finished:
         try:
-            f.write(u.read())
-            finished = True
+            chunk = u.read(CHUNK_SIZE)
+            if chunk:
+                f.write(chunk)
+            else:
+                finished = True
         except:
```
emanueldima: No, this is an infinite loop if the read fails (e.g. url is not accessible). Except must catch a meaningful error, or set finished on true on exception.

demo/b2share_demo/helpers.py
```diff
@@ -418,8 +419,12 @@ def _save_file(url, filename):
...
             current_app.logger.debug("IncompleteRead? %d" % counter)
             counter += counter
```
emanueldima: This counter does nothing now, does it?

emanueldima:
closed, the necessary changes will come in another PR

PR #1183, opened 2016-11-04
Fix file upload documentation

emanueldima:
The content-type seems to be extremely important and when missing will cause the request failure. It's important to document this.


nharraud:
LGTM

PR #1196, opened 2016-11-14
global: oauth and docker fixes

### LINKED ISSUE #1193: gobal: impossible to login twice as the same user with B2ACCESS
**Problem:**
An error is now raised each time one does the following actions: login, logout, login with the same account.
The error happens as B2Share tries to insert for a second time the user in the database. It of course fails as there is already another user with the same email address.

**Investigation:**
It seems that the format returned by `b2share.modules.oauthclient.b2access:account_info` is not valid. It returns `{'nickname':..., 'email': ...}` instead of [this format](https://github.com/inveniosoftware/invenio-oauthclient/blob/90083a55fe52a0678afd86e5e8cb982dae3ffbb3/invenio_oauthclient/contrib/github.py#L134).
It is in fact strange that this was working before.

### LINKED ISSUE #1194: deployment: docker container is not initialized properly
**Problem:**
`b2share.sh` is not ran since [this commit](https://github.com/EUDAT-B2SHARE/b2share/commit/03ef0c3eb4f6b6d40ee16a394af32a634c6698bd#diff-fd4f0e754e6d3f5b1235c30b92df6b18L19). This script was initializing B2Share. Thus the current docker image fails.

### LINKED ISSUE #1195: deployment: docker-compose script does not pin services versions
**Problem:**
B2Share's docker-compose.yml script does not pin the version of elaticsearch, redis, rabbitmq and nginx services used. This could trigger a failed deployment if a new version of one of these services is released with non backward compatible changes.

nharraud:
* FIX Fixes OAuth authentication. (closes #1193)

* Improves oauth authentication by saving B2Access's
  persistent identifier.

* FIX Fixes docker container is not initialization. (closes #1194)

* FIX Pins service versions in docker-compose scripts.
  (closes #1195)

### DIFF
--- a/b2share/modules/oauthclient/b2access.py
+++ b/b2share/modules/oauthclient/b2access.py
@@ -36,7 +36,8 @@ from invenio_oauthclient.handlers import response_token_setter, token_getter, \
 from invenio_oauthclient.proxies import current_oauthclient
 from invenio_oauthclient.signals import account_info_received, \
     account_setup_received
-from invenio_oauthclient.utils import oauth_authenticate, oauth_get_user
+from invenio_oauthclient.utils import oauth_authenticate, oauth_get_user, \
+    oauth_link_external_id
 
 
 REMOTE_APP = dict(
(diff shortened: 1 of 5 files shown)


nharraud:
This PR is marked as WIP because I didn't get yet the time to test the docker container properly and it is not tested by travis.

thijscobben:
I did not install but the changes seem limited and reasonable.

PR #1200, opened 2016-11-17
WIP - community specific data validation

thijscobben:
Hi,

This implements validation of the community specific data against the latest version of all blockschemas belonging to a communityschema.
There was already data validation in the front-end: should it be refactored/removed?

I implemented this in the programming api.py , commit method. But I noticed that the HTTP API view 'patch' (and 'put'?) do not seem to call any of the api methods? But I think Nicholas is also working on this issue in the community administrator / publication workflow PR ?

Curious to hear feedback on this work. It was still relatively hard for me to digest how things work in this part of b2share.

### DIFF
--- a/b2share/modules/deposit/api.py
+++ b/b2share/modules/deposit/api.py
@@ -26,6 +26,7 @@
 import uuid
 from contextlib import contextmanager
 from enum import Enum
+from jsonschema import validate
 from urllib.parse import urlparse, urlunparse
 
 from flask import url_for, g
@@ -42,6 +43,7 @@ from .errors import InvalidDepositDataError, InvalidDepositStateError
 from invenio_records.errors import MissingModelError
 from b2share.modules.records.errors import InvalidRecordError
 from b2share.modules.access.policies import is_under_embargo
+from b2share.modules.schemas import BlockSchema #imports from api
 
 
 class PublicationStates(Enum):
@@ -117,6 +119,7 @@ class Deposit(DepositRecord):
             schema,
             _external=True
         )
+                
         deposit = super(Deposit, cls).create(data, id_=id_)
 
         # create file bucket
@@ -154,7 +157,9 @@ class Deposit(DepositRecord):
                 'Cannot submit a deposit in {} state'.format(
                     self.model.json['publication_state'])
             )
-
+            
+        self.validate_community_specific_schemas()
+        
         # publish the deposition if needed
         if (self['publication_state'] == PublicationStates.published.name
                 # check invenio-deposit status so that we do not loop
@@ -170,6 +175,7 @@ class Deposit(DepositRecord):
         return self
 
     def publish(self, pid=None, id_=None):
+        #throws an exception if not valid
         self['publication_state'] = PublicationStates.published.name
         return super(Deposit, self).publish(pid=pid, id_=id_)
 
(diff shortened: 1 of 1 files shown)


nharraud:
Thanks for the PR, however in my humble opinion there is a better solution. Let me explain why.

*A. There is a simpler solution*

The current JSON Schema could already validate the record if there wasn't one mistake in it:[`"community_specific"` is not in the list of `required` fields](https://github.com/EUDAT-B2SHARE/b2share/blob/07075a3949deb277d2700393438e428a64cda0ea/b2share/modules/schemas/root_schemas/root_schema_v0.json#L144)

By fixing this it should make the record validate properly the record whenever it is created or modified. The code validating a record with the JSON Schema is [here](https://github.com/inveniosoftware/invenio-records/blob/091fca39595339cd4ee35ad111f7be17aa9c10cd/invenio_records/api.py#L266). The `commit` method is called each time the record is modified.

*B. The code suggested in this PR would also create another bug.*

It would validate with the last version of the block schemas whereas the record should use the ones provided in the record's specific schema (the one available at the time the record was created).


**Even with the solution suggested in "A" there will be two side effects**

The first side effect is that the UI doesn't seem to generate a valid record, I don't think that it sets `community_specific` by default. Thus the UI will fail after the fix.

The second side effect is that the draft records won't be valid anymore when they are saved with only some of the required fields set. There are two ways of solving this issue:

- **1.** Modifying the JSON-Schema used for draft records. This JSON Schema is generated [here](https://github.com/EUDAT-B2SHARE/b2share/blob/07075a3949deb277d2700393438e428a64cda0ea/b2share/modules/schemas/serializers.py#L135). It just redirects to the published records' JSON Schema which is generated [here](https://github.com/EUDAT-B2SHARE/b2share/blob/07075a3949deb277d2700393438e428a64cda0ea/b2share/modules/schemas/serializers.py#L134). We would need to generate another schema based on the one for published records but without the constraints on `community_specific` subfields. Those constraints include at least ["minItems", "maxItems"](https://spacetelescope.github.io/understanding-json-schema/reference/array.html#length) and "required".

- **2.** Providing a format checker to `Record.create` and `commit`, see [this](https://github.com/inveniosoftware/invenio-records/blob/091fca39595339cd4ee35ad111f7be17aa9c10cd/invenio_records/api.py#L83). We will have however to wait for PR inveniosoftware/invenio-records#163 to be released before it can work.

**Pros/Cons:**

Solution 1 enables us to have a JSON-Schema visible to outside which really validates drafts. The message to user is clear.

Solution 2 might be easier but it's not sure.

**Implementation advice for solution 1**
It's possible to use [`resolve_references`](https://github.com/inveniosoftware/doschema/blob/a646c936ff206afc37124b0f667dd6d6e8866479/doschema/transform.py#L33) on the schema generated [here](https://github.com/EUDAT-B2SHARE/b2share/blob/07075a3949deb277d2700393438e428a64cda0ea/b2share/modules/schemas/serializers.py#L135). It will resolve all the `$ref` and a return a fully complete schema. After that the transformation should be trivial. However it might be resource consuming to do this every time. We should cache the result or precompute every draft JSON schema and save it somewhere.

** Explanations on how records are validated in Invenio and B2Share**

I already explained above that `Record.validate` validates the record with its JSON-Schema. Now for B2Share this JSON Schema is stored in multiple chunks, root schemas on one side and block schemas on the other. Blocks and Root schemas are referenced in the [community schema](https://github.com/EUDAT-B2SHARE/b2share/blob/07075a3949deb277d2700393438e428a64cda0ea/b2share/modules/schemas/models.py#L110). When a record is created it references the last version of its community schema. When the community publishes a new version of its community schema the existing records still point to the previous versions in order to remain valid.

The community schema [is *compiled* here](https://github.com/EUDAT-B2SHARE/b2share/blob/07075a3949deb277d2700393438e428a64cda0ea/b2share/modules/schemas/api.py#L565). The method `build_json_schema` combines the root and block schema in one schema. We could add here a method `build_draft_json_schema` for solution 1.

emanueldima:
we'll take another path, closing

PR #1204, opened 2016-11-28
Add short rest api migration paragraph

emanueldima:


nharraud:
Looks good to me but I think that there is one thing missing: a link to a testing API so that they can test before putting their code in production. Otherwise we will have a bunch of "Test" records going in production.

emanueldima:
Agree, we have a system for this purpose in v1 but none for v2:

https://github.com/EUDAT-B2SHARE/b2share/issues/1205

PR #1214, opened 2016-12-08
Migrator - CSC specific migration of v1 records

thijscobben:
This should migrate the records from V1 b2share.eudat.eu to any environment.

### DIFF
--- a/demo/b2share_demo/cli.py
+++ b/demo/b2share_demo/cli.py
@@ -25,9 +25,11 @@
 
 from __future__ import absolute_import, print_function
 
+import logging
 import os
 from shutil import rmtree
 import pathlib
+from urllib.parse import urlunsplit
 from shutil import copyfile
 
 import click
@@ -106,6 +108,8 @@ def import_v1_data(verbose, download, token,
          download_directory,limit):
     if verbose:
         click.secho("Importing data to the current instance")
+        logger = logging.getLogger("sqlalchemy.engine")
+        logger.setLevel(logging.ERROR)
     if os.path.isdir(download_directory):
         os.chdir(download_directory)
     else:
@@ -113,12 +117,12 @@ def import_v1_data(verbose, download, token,
     if limit and not download:
         raise click.ClickException("Limit can only be set with download")
     if download:
-        filelist = os.listdir('.')
-        if len(filelist)>0:
-            raise click.ClickException("""You set download_dir to %s .
-            If you want to download files, download_dir should be an empty
-             directory.\n Please empty directory and try again.""" %
-             download_directory)
+        #filelist = os.listdir('.')
+        #if len(filelist)>0:
+        #    raise click.ClickException("""You set download_dir to %s .
+        #    If you want to download files, download_dir should be an empty
+        #     directory.\n Please empty directory and try again.""" %
+        #     download_directory)
         if verbose:
             click.secho("----------")
             click.secho("Downloading data into directory %s" %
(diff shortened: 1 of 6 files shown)


demo/b2share_demo/data/communities/euon.json
```diff
@@ -41,6 +41,7 @@
                             "type": "array",
                             "items": {
                                 "enum": [
+									"english",
```
emanueldima: One of this thing is not like the others 😄 . I'd map English to "Other"

demo/b2share_demo/data/communities/euon.json
```diff
@@ -81,7 +85,8 @@
                                     "OntoStudio",
                                     "KAON",
                                     "PoolParty Thesaurus Server",
-                                    "Other..."
+                                    "Other...",
+									""
```
emanueldima: Empty entry is not needed if we have the "other"

demo/b2share_demo/cli.py
```diff
@@ -106,19 +108,21 @@ def import_v1_data(verbose, download, token,
...
     if limit and not download:
         raise click.ClickException("Limit can only be set with download")
     if download:
-        filelist = os.listdir('.')
-        if len(filelist)>0:
-            raise click.ClickException("""You set download_dir to %s .
-            If you want to download files, download_dir should be an empty
-             directory.\n Please empty directory and try again.""" %
-             download_directory)
+        #filelist = os.listdir('.')
```
emanueldima: should be uncommented?

demo/b2share_demo/helpers.py
```diff
@@ -328,12 +344,20 @@ def download_v1_record(directory, record, verbose=False):
                 click.secho('    Download file "{}"'.format(file_dict.get('name')))
             filepath = os.path.join(directory, 'file_{}'.format(index))
             _save_file(file_dict['url'], filepath)
-            if int(os.path.getsize(filepath)) != int(file_dict.get('size')):
-                raise Exception("downloaded file size differs, {}".format(filepath))
+            #if int(os.path.getsize(filepath)) != int(file_dict.get('size')):
```
emanueldima: But we must make sure that the sizes are the same! It must be exactly the same data.

demo/b2share_demo/helpers.py
```diff
@@ -408,24 +451,62 @@ def _process_record(rec):
...
-            finished = True
+            chunk = u.read(CHUNK_SIZE)
+            if chunk:
+                f.write(chunk)
+            else:
+                finished = True
         except:
-            current_app.logger.debug("IncompleteRead? %d" % counter)
-            counter += counter
+            finished = True
```
emanueldima: incorrect to assume that the file is correctly finished downloading, I'd simply quit with error message

demo/b2share_demo/data/communities/euon.json
thijscobben: This is required to pass validation on usedOntologyEngineeringTool="" which appears and to my opinion we cannot map the case "No tool used" to "Other tool used" ? So I keep the value. Note that usedOntologyEngineeringTool is not a required field in this schema, so we can assume that "empty" indeed means "No tool used" (as would leaving out the key/value all together imply too).

demo/b2share_demo/data/communities/euon.json
thijscobben: "english" was a value specified by one of the communities I advice to retain it in the schema this is their choice.

demo/b2share_demo/helpers.py
thijscobben: fixed. This also voids the case you mention below:

demo/b2share_demo/helpers.py
thijscobben: Fixed by uncommenting file size check.

Issue #1205, opened 2016-11-28
Make clear the purpose of an B2SHARE instance


emanueldima:
Some users confuse the production server and the testing ones, and create many test records on production instances. To avoid this we should have clear messages about the purpose of the instance in the following locations:

1. the home page
2. the depositing page
3. the http api documentation

On production systems, these messages should provide links to testing instances.

PR #1215, opened 2016-12-08
Add api root; display ui site function; fix missing node package

emanueldima:
fixes https://github.com/EUDAT-B2SHARE/b2share/issues/1205


b2share/config.py
```diff
@@ -260,3 +260,5 @@
 PREFERRED_URL_SCHEME = 'http'
 
 FILES_REST_DEFAULT_MAX_FILE_SIZE = 1000 * 1024 * 1024 # 1 GB by default for RC4
+
+SITE_FUNCTION = 'testing'
```
nharraud: I don't see any other place where `SITE_FUNCTION`. It seems that this would mark the docker container as `testing` too. @llehtine will then need to override this when deploying in production.

INSTALL.rst
```diff
@@ -33,11 +33,17 @@ account:
...
+1.2 Prerequisite: configuring B2SHARE
+-------------------------------------
+
+B2SHARE has a long list of configuration parameters, with default values
+available for inspection in the `b2share/config.py` file.
+
+After B2SHARE is provisioned (after the b2share.sh script runs
+``/usr/bin/b2share demo load_config``), local configurations can be created by
+adding the same parameters (with different values) in the
+``/usr/var/b2share-instance/b2share.cfg`` file.
```
nharraud: The configuration can be overriden with environment variables when starting B2Share.
This can be done with the docker container too.
The format of the environment variable is `B2SHARE_<CONFIGURATION_VARIABLE_NAME>`

b2share/config.py
```diff
@@ -261,4 +261,9 @@
 
 FILES_REST_DEFAULT_MAX_FILE_SIZE = 1000 * 1024 * 1024 # 1 GB by default for RC4
 
-SITE_FUNCTION = 'testing'
+# prominently displayed on the front page, except when set to "production"
```
nharraud: and also returned by the REST API when querying `http://<HOSTNAME>/api`

b2share/config.py
```diff
@@ -261,4 +261,9 @@
...
+SITE_FUNCTION = 'demo' # set to "production" on production instances
+
+# if the TRAINING_SITE_LINK parameter is not empty, a message will show up
+# on the front page redirecting the testers to this link
+TRAINING_SITE_LINK = "http://trng-b2share.eudat.eu"
```
nharraud: We need to be careful with this one. It is specific to CSC. It will be displayed on demo instances people start themselves.

It would be more appropriate to have a customizable header/footer or a "flash" message shown when the user goes to the front page.

webui/src/docs/rest-api.jsx
```diff
@@ -139,6 +139,9 @@ module.exports = function() {
                 training site. Use this URL for testing.
             </li>
         </ul>
+        <p>Please make sure that you are not using production instances for
+            creating test records or testing the API in general.
```
nharraud: We should provide or at least recommend a standardized value for "site_function" when B2Share is in production. That way people can develop a standard client library and check for the value.

INSTALL.rst
emanueldima: That's new for me, good to know.

nharraud: That's what happens when we use `B2SHARE_JSONSCHEMAS_HOST=...`. It overrides the `JSONSCHEMAS_HOST` config variable.

b2share/config.py
emanueldima: Ok, might be better to make this one empty by default.

PR #1224, opened 2016-12-13
Update migrator version

thijscobben:
Conforms to the OpenAIRE root schema


demo/b2share_demo/helpers.py
```diff
@@ -441,8 +457,12 @@ def _process_record(rec):
     else:
         return None
     if 'PID' in rec.keys():
-        result['alternate_identifier'] = rec['PID']
-
+        result['alternate_identifiers'] = [
+            {'alternate_identifier_type':'EPIC_PID',
```
emanueldima: please use `ePIC_PID` here, this identifier type is also used elsewhere.

PR #1228, opened 2016-12-14
Update root schema, serializers, and migration code

emanueldima:


demo/b2share_demo/helpers.py
```diff
@@ -455,14 +466,15 @@ def _process_record(rec):
         community = Community.get(name='CLARIN')
         result['community']=str(community.id)
     else:
-        return None
+        raise Exception("Community not found for domain: `{}`".format(rec['domain']))
```
thijscobben: Are we sure we want to raise an exception here? Not just log the issue and continue?

b2share/modules/records/serializers/schemas/marcxml.py
```diff
@@ -37,39 +37,62 @@ class RecordSchemaMarcXMLV1(Schema):
 
     def get_id(self, obj):
         pids = obj['metadata'].get('_pid')
-        p = [p['value'] for p in pids if p['type'] == 'b2share_record']
+        p = [p['value'] for p in pids if p['type'] == 'b2rec']
```
emanueldima: This is just for compatibility with v1, but we can change it.

b2share/modules/records/serializers/schemas/datacite.py
```diff
@@ -76,43 +68,57 @@ class DataCiteSchemaV1(Schema):
...
         fields.Nested(AlternateIdentifierSchema, attribute='metadata._pid'))
-    rightsList = fields.List(fields.Function(
-        lambda o: {'rights': o['metadata'].get('licence')}))
+    rightsList = fields.Method('get_rights')
     descriptions = fields.Method('get_descriptions')
 
     def get_creators(self, obj):
-        crs = obj['metadata'].get('creators', ['[Unknown]'])
-        return [{'creatorName':c} for c in crs]
+        crs = obj['metadata'].get('creators', [{'creator_name':'[Unknown]'}])
```
emanueldima: Datacite requires at least one creator name, the xml is not valid otherwise. If we make the DOI optional we can remove this hack.

nharraud: Then I would add it at the serialization time in datacite format. Datacite doesn't enforce completely our JSON.

nharraud: ok now I understand better. Thanks

PR #1237, opened 2016-12-15
WIP one-time-only generate_pid_migrator command

thijscobben:
generates a curl command for each record in its' instance to update the url of the PID from the v1 URL to the v2 URL

### DIFF
--- a/demo/b2share_demo/cli.py
+++ b/demo/b2share_demo/cli.py
@@ -163,12 +163,16 @@ def generate_pid_migrator(base_url):
     response = requests.get(url,params)
     recs = json.loads(response.text)['hits']['hits']
     for rec in recs:
-        url_value = rec['links']['self']
-        alt_ids = rec['alternate_identifiers']
-        epic_url = None
-        for aid in alt_ids:
-            if aid['alternate_identifier_type'] == 'ePIC_PID':
-                epic_url = aid['alternate_identifier']
-        if not(epic_url is None):
-            print("TEST %s %s" % (epic_url, url_value))
+        url_value = rec['links']['self'].replace("/api/records","/records")
+        if 'alternate_identifiers' in rec['metadata'].keys():
+            alt_ids = rec['metadata']['alternate_identifiers']
+            epic_url = None
+            for aid in alt_ids:
+                if aid['alternate_identifier_type'] == 'ePIC_PID':
+                    epic_url = aid['alternate_identifier']
+            if not(epic_url is None):
+                curl_comm = "curl -X PUT -v -H 'Accept:application/json'"
+                curl_data = '{"type":"URL","parsed_data":"%s"}' % url_value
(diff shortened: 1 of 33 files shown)


demo/b2share_demo/helpers.py
```diff
@@ -315,7 +315,8 @@ def download_v1_data(token, target_dir, limit=None, verbose=False):
         if len(recs) == 0:
             return # no more records
         for record in recs:
-            if record.get('files') == 'RESTRICTED':
+            if (record.get('files') == 'RESTRICTED' or 
+                not(record.get('open_access')) ):
```
emanueldima: I disagree with this change, we should at least try to make the record and migrate the files afterward, in other ways

demo/b2share_demo/cli.py
```diff
@@ -147,3 +149,28 @@ def import_v1_data(verbose, download, token,
...
+    for rec in recs:
+        url_value = rec['links']['self'].replace("/api/records","/records")
+        if 'alternate_identifiers' in rec['metadata'].keys():
+            alt_ids = rec['metadata']['alternate_identifiers']
+            epic_url = None
+            for aid in alt_ids:
+                if aid['alternate_identifier_type'] == 'ePIC_PID':
+                    epic_url = aid['alternate_identifier']
+            if not(epic_url is None):
+                curl_comm = "curl -X PUT -v -H 'Accept:application/json'"
```
emanueldima: a space at the end would be needed

demo/b2share_demo/cli.py
```diff
@@ -147,3 +149,28 @@ def import_v1_data(verbose, download, token,
...
+                curl_data = '{"type":"URL","parsed_data":"%s"}' % url_value
+                curl_comm += "--data ='[%s] %s" % (curl_data, epic_url)
```
emanueldima: We also need to add the PID server authentication, taken from the configuration file:
````
# CFG_EPIC_USERNAME
# CFG_EPIC_PASSWORD
# CFG_EPIC_BASEURL = 'https://epic4.storage.surfsara.nl/v2_A/handles/'
````

so add `'-u {}:{} '.format(username, pass)'`

File: demo/b2share_demo/cli.py [hunk 4]
```diff
@@ -147,3 +149,28 @@ def import_v1_data(verbose, download, token,
...
+    params = {}
+    params['size'] = 1000
+    params['page'] = 1
+    response = requests.get(url,params)
+    recs = json.loads(response.text)['hits']['hits']
+    for rec in recs:
+        url_value = rec['links']['self'].replace("/api/records","/records")
+        if 'alternate_identifiers' in rec['metadata'].keys():
+            alt_ids = rec['metadata']['alternate_identifiers']
+            epic_url = None
```
emanueldima: The url where the PID can be changed is not identical to the global handle URL. We must take the suffix part (UUID) and place it at the end of our own handle server (see `config['CFG_EPIC_BASEURL']`)

Issue #1172, opened 2016-10-25
api: validate record's REST API PATCH and PUT inputs


nharraud:
**Problem:**
A record's JSON Schema can only check the resulting record structure and values. It does not check what if a modification is allowed.
We need to add some mechanism to forbid changes.

Here is an incomplete list of modifications which should not be allowed for a `draft`:
- [x] change the 'publication_state' to an invalid state
  - [x] change from `draft` to `published` or `submitted` when the community workflow does not allow it.
  - [x] change to invalid values. This currently already handled by the root JSON Schema but should be moved somewhere else if communities need to create completely custom workflows.
- [x] the community ID of a record should never be modified. Its schema depends on it.
- [x] internal fields should not be modified from outside. Those fields include:
  - `_pid`
  - `_files`
  - `_oai`
  - `_deposit`
  - `version`
- [x] inconsistent `embargo_date` and `open_access`. But this could be handled by the JSON Schema. This is currently automatically fixed, but a 400 error should be instead sent when a user explicitly asks for `open_access = True` and `embargo_date` in the future.

A published record has all the same constraint except for `publication_state` which should always be `published`.

emanueldima:
@nharraud Can we close this one?

nharraud:
@emanueldima Just the last one could be improved. Removing urgent tag.

thijscobben:
Fixed by celery task that makes combination of open_access and future embargo_date consistent.

PR #1240, opened 2016-12-16
api: REST API input validation

nharraud:
* Prevents records immutable fields from being modified using the
  REST API PATCH method. (addresses #1172)

* Disables record REST API PUT method.

### DIFF
--- a/b2share/modules/deposit/loaders.py
+++ b/b2share/modules/deposit/loaders.py
@@ -21,28 +21,35 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-"""B2share records triggers."""
+"""B2Share deposit input loaders."""
 
-from __future__ import absolute_import, print_function
+from flask import abort, request
+from invenio_rest.errors import RESTValidationError, FieldError
 
-import pytz
-from invenio_records.signals import before_record_update
-from invenio_rest.errors import FieldError
-from .errors import AlteredRecordError
+IMMUTABLE_PATHS= {
+    # fields added by the serializer
+    '/owners'
+    '/ePIC_PID'
+    '/DOI',
+    '/files',
+    # real fields
+    '/community',
+    '/$schema',
+    '/_pid',
+    '/_oai',
+    '/_files',
+    '/_deposit',
+}
 
 
-def register_triggers(app):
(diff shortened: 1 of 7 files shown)


b2share/modules/deposit/loaders.py
```diff
@@ -0,0 +1,55 @@
...
+    '/ePID_PID'
+    '/DOI',
+    # real fields
+    '/community',
+    '/$schema',
+    '/version',
+    '/_pid',
+    '/_files',
+    '/_deposit'
+    '/creators',
```
nharraud: Sorry I thought about owners, not creators. I'll fix it today and rebase the PR.

nharraud: I was wondering the same thing about "version". For now it is a field provided by the user so it would make sense to let the user change it, no?

emanueldima: Yes, I've also wondered if we can remove it and use the db version instead: https://github.com/EUDAT-B2SHARE/b2share/issues/1242

nharraud: Maybe I misunderstood the meaning of this `version` field then. I thought that it was some outside version number. It seems to be the case in V1.
If it is the version given by B2Share/Invenio then we should completely remove the field and provide the version number later at the same level as "_created" and "_updated", i.e. outside of the metadata.

nharraud: If we remove it what do we do with the user-provided version number in V1? Example: https://b2share.eudat.eu/record/580

nharraud: I suggest:
- keeping the version field in the current root schema as we have to keep what exists in v1 and some users used this field. Enable also any modification to this field as it is user-managed.
- create another version of the root schema without the version field before releasing B2Share v2 in February. That way every new record will not have the user-managed version field.
- before releasing B2Share v2 software, serialize the invenio version number in the record output of the REST API.

nharraud: Updated the list:
- removed "version" and "creators"
- fixed "ePIC_PID" which was misspelled
- added "files" which is generated by the serializer and "_oai"

I have some doubts about "publication_date", looks like we could remove this field later in a new root schema. As we already have "created" date and the embargo_date it is just a duplicate. But we can keep it in root schema v0 just to simplify the migration.

PR #1244, opened 2016-12-19
Make DOI assignment configurable

emanueldima:


b2share/modules/records/serializers/schemas/datacite.py
```diff
@@ -35,7 +35,7 @@ class IdentifierSchema(Schema):
 
     def get_doi(self, pids):
         p = [p['value'] for p in pids if p['type'] == 'DOI_RESERVED']
-        return str(p[0])
+        return str(p[0]) if p else ""
```
nharraud: IMHO `None` would be more appropriate than `""`.

emanueldima: I tried, the serializer crashes...

nharraud: I have [found this](https://guidelines.openaire.eu/en/latest/data/use_of_datacite.html#what-s-different)
```
OpenAIRE accepts other persistent identifier schemes and not only a DOI in 1.1 identifierType (M).
```
Maybe we could use ePIC PIDs with PID type "Handle".
See also [the list of PID types accepted by OpenAIRE](https://guidelines.openaire.eu/en/latest/data/field_identifier.html#d-identifiertype)

emanueldima: Fine, but this particular serializer is for datacite, where DOIs are mandatory.

PR #1246, opened 2016-12-19
Generate ePIC_PID migrator script

thijscobben:
It's a one-time only thing. Code should be removed after we used it.


demo/b2share_demo/cli.py
```diff
@@ -142,8 +145,38 @@ def import_v1_data(verbose, download, token,
...
+            alt_ids = rec['metadata']['alternate_identifiers']
+            epic_url = None
+            for aid in alt_ids:
+                if aid['alternate_identifier_type'] == 'ePIC_PID':
+                    handle_url = aid['alternate_identifier']
+                    epic_pid = handle_url.rsplit("/",1)[-1]
+                    epic_url = epic_base_url + epic_pid  
+            if not(epic_url is None):
+                curl_comm = "curl -X PUT -v -H 'Accept:application/json' "
+                curl_comm += "-d %s:%s " % (epic_username, epic_password)
```
emanueldima: I think it's `-u` here

PR #1247, opened 2016-12-19
api: access request and abuse report fix

### LINKED ISSUE #1239: api: refactor access request and abuse notification in records REST API
**Problem:**
The [current implementation of `RecordsAbuseResource` and `RequestAccessResource`](https://github.com/EUDAT-B2SHARE/b2share/blob/6c3ad0b9fcbf6a1a79ada6806b56bb04a9dbbf1c/b2share/modules/records/views.py#L359) is completely flawed as it subclasses `RecordResource` class without being at all a Record endpoint.
This needs to be reimplemented.

### LINKED ISSUE #1245: api: record access requests are sending email to support
**Problem:**
Current implementation of Record Access requests are sending emails to support.
They should instead send an email to 'contact_email' or, if it is not set, to the record owners.

nharraud:
* FIX Sends email to contact_email or owners instead of sending it
  to support. (closes #1239)

* FIX RecordsAbuseResource and RequestAccessResource are now
  subclasses of ContentNegotiatedMethodView instead of
  RecordResource. (closes #1245)

### DIFF
--- a/b2share/modules/records/views.py
+++ b/b2share/modules/records/views.py
@@ -34,7 +34,8 @@ from invenio_records_files.api import Record
 from invenio_rest.errors import RESTValidationError
 from invenio_search import RecordsSearch
 from jsonschema.exceptions import ValidationError
-from invenio_records_rest.views import (RecordsListResource, RecordResource,
+from invenio_records_rest.views import (pass_record,
+                                        RecordsListResource, RecordResource,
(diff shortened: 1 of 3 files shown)


nharraud:
@emanueldima @SarahBA I didn't test with the UI yet, but the input didn't change. I will test it now.

PR #1249, opened 2016-12-19
Update http api docs to conform to new root schema

emanueldima:


webui/src/main.jsx
```diff
@@ -155,7 +155,7 @@ const Footer = React.createClass({
                         </ul>
                         <ul className="list-inline pull-right">
                             <li><a href="/help/terms-of-use">Terms of Use</a></li>
-                            <li><a href="/help/api">Rest API</a></li>
+                            <li><a href="/help/api">HTTP API</a></li>
```
nharraud: *REST* is a term which is now commonly used in documentation, see [Google](https://developers.google.com/fit/rest/), [Zenodo](https://developer.paypal.com/docs/api/) or [Paypal](https://developer.paypal.com/docs/api/)
But I don't see any issue with calling it `HTTP API`, so why not.

PR #1288, opened 2017-01-04
PID swapper

thijscobben:
Swaps _pid/ePIC_PID (the one we wrongfully created during migration) with alternate_identifiers/ePIC_PID (the old one from v1)

Note this creates a new version of the involved records that we cannot see.
This can currently only be tested/seen from the b2share shell.

### DIFF
--- a/demo/b2share_demo/cli.py
+++ b/demo/b2share_demo/cli.py
@@ -41,6 +41,7 @@ from flask import current_app
 from invenio_db import db
 from invenio_files_rest.models import Location
 from invenio_indexer.api import RecordIndexer
+from invenio_records.api import Record
 
 from .helpers import load_demo_data, download_v1_data, process_v1_record
 from . import config as demo_config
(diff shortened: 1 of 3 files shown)


demo/b2share_demo/cli.py
```diff
@@ -236,6 +237,50 @@ def is_same_url(url1, url2):
...
+            found = False
+            for aid in aids:
+                if aid['alternate_identifier_type']=='ePIC_PID':
+                    new_pid = aid['alternate_identifier']
+                    _pid = inv_record['_pid']
+                    for pid in _pid:
+                        if pid['type']=='ePIC_PID':
+                            old_pid = pid['value']
+                            found = True
+            if not found:
```
emanueldima: `found` is true only for the imported records, but there might still be some not imported (new), and in this case the script will show unnecessary errors

demo/b2share_demo/cli.py
```diff
@@ -236,6 +237,50 @@ def is_same_url(url1, url2):
...
+                error_msg = """***** ERROR - this record does not have ePIC_PID 
+                    in alternate_identifiers"""
```
emanueldima: incorrect message, see when found becomes True

demo/b2share_demo/cli.py
```diff
@@ -236,6 +237,50 @@ def is_same_url(url1, url2):
...
+            aids = inv_record['alternate_identifiers']
+            found = False
+            for aid in aids:
+                if aid['alternate_identifier_type']=='ePIC_PID':
+                    new_pid = aid['alternate_identifier']
+                    _pid = inv_record['_pid']
+                    for pid in _pid:
+                        if pid['type']=='ePIC_PID':
+                            old_pid = pid['value']
+                            found = True
```
emanueldima: after this line there should be two 'break' statement: it's possible there are multiple PIDs, but only the first instance should be considered

demo/b2share_demo/cli.py
```diff
@@ -236,6 +237,50 @@ def is_same_url(url1, url2):
...
+                    in alternate_identifiers"""
+                print(error_msg)
+                print(inv_record['titles'])
+                print(rec['id'])
+                print("********")
+            else:
+                print("SWAPPING %s %s" % (old_pid, new_pid))
+                for pid in inv_record['_pid']:
+                    if pid['type']=='ePIC_PID':
+                        pid['value']=new_pid
```
emanueldima: break after this line, otherwise we overwrite all PIDs with the same value

demo/b2share_demo/cli.py
```diff
@@ -236,6 +237,50 @@ def is_same_url(url1, url2):
...
+                for aid in inv_record['alternate_identifiers']:
+                    if aid['alternate_identifier_type']=='ePIC_PID':
+                        aid['alternate_identifier']=old_pid
+                inv_record.commit()
```
emanueldima: break after this line, otherwise we overwrite all PIDs with the same value

Issue #1279, opened 2016-12-22
New PIDs point to incorrect URL


emanueldima:
When publishing a new record the generated PID does not point to the record's url, but to a url build using the record's uuid. The result for the user is a 404 not found error.

PR #1289, opened 2017-01-04
Fix PID and DOI assignation

emanueldima:
fixes https://github.com/EUDAT-B2SHARE/b2share/issues/1279

### DIFF
--- a/b2share/modules/records/minters.py
+++ b/b2share/modules/records/minters.py
@@ -41,61 +41,61 @@ from .b2share_epic import createHandle
 
 def b2share_record_uuid_minter(record_uuid, data):
     """Use record's UUID as unique identifier."""
-    provider = RecordUUIDProvider.create(
-        object_type='rec', object_uuid=record_uuid,
+    rec_pid = RecordUUIDProvider.create(
+        object_type='rec',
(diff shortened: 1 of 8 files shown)


b2share/modules/records/minters.py
```diff
@@ -41,111 +41,110 @@
...
-    b2share_oaiid_minter(record_uuid, data)
-    b2share_pid_minter(record_uuid, data)
-    if current_app.config.get('AUTOMATICALLY_ASSIGN_DOI'):
-        b2share_doi_minter(record_uuid, data)
+    b2share_oaiid_minter(rec_pid, data)
+    b2share_pid_minter(rec_pid, data)
+    if current_app.config.get('TESTING', False) or current_app.config.get('FAKE_DOI', False):
+        b2share_doi_minter(rec_pid, data, fake_it=True)
+    elif current_app.config.get('AUTOMATICALLY_ASSIGN_DOI'):
+        b2share_doi_minter(rec_pid, data)
```
thijscobben: So in the case FAKE_DOI=True and AUTOMATICALLY_ASSIGN_DOI=False we assign (fake) DOI. Does that make sense?

Issue #1285, opened 2017-01-02
remove migration code duplication


emanueldima:
the following code duplication should be fixed: 
- [ ] `b2share_demo/helpers.py:_process_record` vs. `b2share_demo/migration.py:convert_old_record_to_new`
- [ ] `b2share_demo/helpers.py:_match_community_specific_metadata` vs. `b2share_demo/migration.py:_match_community_specific_metadata`

PR #1293, opened 2017-01-05
Migration updates: cli to list alternate ids, cleaning up

emanueldima:
fixes https://github.com/EUDAT-B2SHARE/b2share/issues/1285

### DIFF
--- a/demo/b2share_demo/cli.py
+++ b/demo/b2share_demo/cli.py
@@ -25,14 +25,9 @@
 
 from __future__ import absolute_import, print_function
 
-import json
-import logging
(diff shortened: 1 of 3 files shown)


demo/b2share_demo/cli.py
```diff
@@ -234,8 +237,39 @@ def is_same_url(url1, url2):
         u1.path == u2.path and u1.query == u2.query
 
 
+
+
+@demo.command()
```
thijscobben: This should now be moved to migration_cli and made a migrate command.

Issue #1162, opened 2016-10-19
DOI errors when importing records from production


emanueldima:
datacite.errors.DataCiteBadRequestError

emanueldima:
for the moment avoided by https://github.com/EUDAT-B2SHARE/b2share/pull/1244

Issue #1256, opened 2016-12-20
CSC: curate B2Share.eudat.eu data before migrating to the final version of B2Share v2


nharraud:
**Problem:**
Apparently there are some invalid data in b2share.eudat.eu. As we upgrade the server we will need for now to loosen some constraints on B2Share V2 release candidates. However we cannot release this in the final version. Thus we need to curate b2share.eudat.eu data before migrating to the final version of B2share v2.

Concerned fields are:
- [ ] `publication_date`: it was discovered as elasticsearch was refusing some records. (See the quick fix in #1255 which should be removed later)
- [ ] `creators`: for some document the array is empty.
- [ ] `publisher`: for some document this field is set to empty string.

hevp:
This is no longer relevant. Issue closed.

PR #1312, opened 2017-01-13
Fix datacite serialization errors

emanueldima:
fixes https://github.com/EUDAT-B2SHARE/b2share/issues/1162

### DIFF
--- a/b2share/modules/records/serializers/schemas/datacite.py
+++ b/b2share/modules/records/serializers/schemas/datacite.py
@@ -58,7 +58,7 @@ class DataCiteSchemaV1(Schema):
     titles = fields.Function(lambda o: o['metadata'].get('titles'))
 
     publisher = fields.Function(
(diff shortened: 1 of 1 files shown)


b2share/modules/records/serializers/schemas/datacite.py
```diff
@@ -72,7 +72,7 @@ class DataCiteSchemaV1(Schema):
     descriptions = fields.Method('get_descriptions')
 
     def get_creators(self, obj):
-        crs = obj['metadata'].get('creators', [{'creator_name':'[Unknown]'}])
+        crs = obj['metadata'].get('creators') or [{'creator_name':'[Unknown]'}]
```
nharraud: Do we have documents with `None` or empty list of creators?

emanueldima: Yes, I found records with an empty list of creators

nharraud: Ok. I added this to the curate issue #1256

PR #1340, opened 2017-01-23
Make datacite serialization target schema v3.1

emanueldima:


b2share/modules/records/serializers/schemas/datacite.py
```diff
@@ -119,6 +119,17 @@ def get_contributors(self, obj):
                 for c in obj['metadata'].get('contributors', [])]
 
     def get_descriptions(self, obj):
-        return [{'description': d['description'],
+        items = []
+        for d in obj['metadata'].get('descriptions', []):
+            d = description_to_datacite_v3(d)
```
nharraud: What about changing that into:
```python
datacite_v3_description_types = set([<ALLOWED TYPES>])
def get_descriptions(self, obj):
    return [
        {
            'description': description['description'],
            'descriptionType': description['description_type']
                if description['description_type'] in datacite_v3_description_types else 'Other',
        }
        for description in obj['metadata'].get('descriptions', [])
    ]
```
It would be more efficient as there are less functions called and no `append` call. You can see [this page](https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Loops) for more details.
The suggested implementation also allows to have more custom types without having to ever change this function.

emanueldima:
@nharraud Better?

b2share/modules/records/serializers/schemas/datacite.py
```diff
@@ -119,6 +119,10 @@ def get_contributors(self, obj):
                 for c in obj['metadata'].get('contributors', [])]
 
     def get_descriptions(self, obj):
+        datacite_v3_description_types = set([
+            "Abstract", "Methods", "SeriesInformation", "TableOfContents"])
```
nharraud: I would put the set creation out of the function. No need to recreate it each time. You can put an underscore first in order to avoid polluting the module's namespace `_datacite_v3_description_types `.

b2share/modules/records/serializers/schemas/datacite.py
```diff
@@ -119,6 +119,10 @@ def get_contributors(self, obj):
...
+        convert = lambda dt: dt if dt in datacite_v3_description_types else "Other"
```
nharraud: The lambda is called only in one place. IMHO we could as well inline it directly in the list comprehension. Calling functions in python is expensive and there is not much difference in code readability.

emanueldima:
How about now?

Issue #1339, opened 2017-01-23
DOI registration failures due to schema versions mismatch


emanueldima:
B2SHARE root metadata schema is based on DataCite v4, but DOI registration requires DataCite v3.1. Record https://b2share.eudat.eu/api/records/a921cfe6422544ec96302f60dece7393 has this problem (`descriptionType: TechnicalInfo` has been added in v4) and needs to have its DOI re-registered.

emanueldima:
see https://github.com/EUDAT-B2SHARE/b2share/pull/1340

emanueldima:
https://github.com/EUDAT-B2SHARE/b2share/pull/1350 will allow us to fix errors in production

emanueldima:
#1350 failed: the internal b2share pid is registered, but the DOI is not. Another approach is needed.

emanueldima:
DOIs correctly allocated on 19 May 2017

PR #1350, opened 2017-01-31
Add cli to register reserved dois with datacite

emanueldima:
needed to resolve issues like https://github.com/EUDAT-B2SHARE/b2share/issues/1339


demo/b2share_demo/helpers.py
```diff
@@ -291,3 +292,13 @@ def community_id_match(match):
         source
     )
 
+
+def list_db_records():
+    """A generator for all the records in the current instance"""
+    query = RecordMetadata.query.filter(RecordMetadata.json != None)
+    for obj in query.all():
+        record = Record(obj.json, model=obj)
+        if record.get('$schema', "").endswith('draft_json_schema'):
```
thijscobben: Why not if record.get("publication_state") is unequal to deposit.PublicationStates.published ?

Issue #1160, opened 2016-10-18
Add 'Find own drafts' option to the user profile and HTTP docs


emanueldima:
To find the drafts of an user:
1. set the ‘drafts=1’ arg to a suitable http search call
2. filter out the drafts based on the status (set q=publication_state:draft)

https://vm0045.kaj.pouta.csc.fi/api/records/?drafts=1&q=publication_state%3Adraft

ALSO, update the rest documentation with the new option for searching in own drafts

emanueldima:
http api documentation updated in https://github.com/EUDAT-B2SHARE/b2share/pull/1169

SarahBA:
fixed by https://github.com/EUDAT-B2SHARE/b2share/pull/1364

PR #1364, opened 2017-02-08
Add 'Find own drafts' option to the user profile

SarahBA:
closing https://github.com/EUDAT-B2SHARE/b2share/issues/1160


emanueldima:
I agree with the changes, but right now the users cannot distinguish between a search result page and a search drafts page, we should change some of the strings on the page to make this distinction clear (like the page title, for instance).

PR #1371, opened 2017-02-13
Add oai sets and cli commands for updating sets and records

emanueldima:


b2share/modules/records/cli.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of EUDAT B2Share.
-# Copyright (C) 2017 University of Tübingen
+# Copyright (C) 2016 University of Tuebingen, CERN, SurfSara
```
nharraud: back to the future?

b2share/modules/oaiserver/cli.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@click.group()
+def oai():
+    """OAI commands."""
+
+
+@oai.command()
+@with_appcontext
+def update_sets():
```
nharraud: we could add a fixme, for removing those commands once the b2share.eudat.eu is migrated.

b2share/modules/oaiserver/cli.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EUDAT B2Share.
+# Copyright (C) 2016 University of Tuebingen, CERN, SurfSara
```
nharraud: back to the future?

b2share/modules/oaiserver/__init__.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EUDAT B2Share.
+# Copyright (C) 2016 University of Tuebingen, CERN.
+# Copyright (C) 2015 University of Tuebingen.
```
nharraud: "Double" back to the future?

b2share/modules/records/utils.py
```diff
@@ -45,3 +43,14 @@ def is_deposit(record):
         bool: True if the record is a deposit record, else False.
     """
     return record.json['$schema'].endswith('#/draft_json_schema')
+
+
+
+
+def list_db_published_records():
+    """A generator for all the published records"""
+    query = RecordMetadata.query.filter(RecordMetadata.json != None)
```
nharraud: We cab make a join with the pid table and filter on the pid type for better performances.

b2share/modules/oaiserver/cli.py
emanueldima: added an issue: https://github.com/EUDAT-B2SHARE/b2share/issues/1373

b2share/modules/records/utils.py
emanueldima: How important is this? Currently this fn is not on a critical code path: https://github.com/EUDAT-B2SHARE/b2share/issues/1374

nharraud: It can wait for now as it is used only in the CLI. However it doubles the number of records returned by the DB as it goes through the deposits too.

PR #1379, opened 2017-02-14
Accounts rest

nharraud:
* New users and roles rest api
* Updates Invenio dependencies.

This PR is waiting for inveniosoftware/invenio-accounts-rest#23 to be merged and the next alpha version of  inveniosoftware/invenio-accounts-rest to be released.


nharraud:
The develop and current builds still fail because of inveniosoftware/invenio-oaiserver#112

b2share/modules/users/views.py
```diff
@@ -37,7 +37,7 @@
 blueprint = Blueprint(
     'b2share_users',
     __name__,
-    url_prefix='/users'
+    url_prefix='/user'
```
emanueldima: This change breaks the UI, please update this file: https://github.com/EUDAT-B2SHARE/b2share/blob/evolution/webui/src/data/server.js#L14 (just the urls should be enough)

b2share/modules/communities/api.py
```diff
@@ -217,7 +215,6 @@ def patch(self, patch):
             'description': self.model.description,
             'logo': self.model.logo,
             'publication_workflow': self.model.publication_workflow,
-            'restricted_submission': self.model.restricted_submission,
```
emanueldima: Is this change intended?

b2share/modules/communities/api.py
```diff
@@ -139,12 +141,8 @@ def create_community(cls, name, description, logo=None, id_=None,
                 kwargs = {}
                 if id_ is not None:
                     kwargs['id'] = id_
-                model = CommunityMetadata(name=name,
-                                          description=description,
-                                          logo=logo,
-                                          publication_workflow=publication_workflow,
```
emanueldima: Is this change intended?

b2share/modules/communities/api.py
```diff
@@ -95,19 +97,19 @@ def get_all(cls, start=None, stop=None, name=None):
         """Searches for matching communities."""
         if (start is None and stop is None):
             if name is None:
-                metadata = CommunityMetadata.query.order_by(CommunityMetadata.created)
+                metadata = CommunityMeta.query.order_by(CommunityMeta.created)
```
nharraud: Oups I think you are right. I have rebased and added this by mistake.

nharraud: Stange that the tests didn't detect it.

b2share/modules/users/views.py
nharraud: ok thanks !

PR #1381, opened 2017-02-16
CLI for set_schema COMMUNITY BLOCK_SCHEMA_VERSION_FILE

thijscobben:
Rudimentary doc in docs/Administration.rst .

PR #1384, opened 2017-02-20
CLI Schema management

emanueldima:
This PR fixes a few issues with https://github.com/EUDAT-B2SHARE/b2share/pull/1381 and builds on it to add CLI commands for creating community specific metadata block schemas.


tests/b2share_functional_tests/test_schema_cli.py
```diff
@@ -0,0 +1,97 @@
...
+        result = runner.invoke(communities_cmd, args, obj=script_info)
+        assert result.exit_code == 0
+
+        with runner.isolated_filesystem():
+            with open("schema.json", "w") as f:
+                f.write(json.dumps(test_schema))
+
+            # initialize the root schema in the test environment
+            result = runner.invoke(schemas_cmd, ['init'], obj=script_info)
+            assert result.exit_code == 0
```
nharraud: Ok for now, but normally the test should also check that the command actually did something. It could return 0 by mistake, not commit at the end or have another issue.

b2share/modules/schemas/cli.py
```diff
@@ -154,124 +153,225 @@ def block_schema_edit(verbose, name, community, deprecated, block_schema_id):
...
-    except json.decoder.JSONDecodeError:
-        raise click.ClickException("%s is not valid JSON" % json_file)
+    with open(json_file, 'r') as f:
+        try:
+            schema_dict = json.load(f)
+        except ValueError: # json decode error is subclass of ValueError
+            raise click.ClickException("%s is not valid JSON" % json_file)
+
     try:
         validate_metadata_schema(schema_dict)
```
nharraud: Why validate the schema here when the API already does it?

b2share/modules/schemas/cli.py
```diff
@@ -65,38 +60,40 @@ def init(verbose):
...
+        click.secho("""Creating block schema with name %s to be maintained by
         community %s""" % (name, community))
     comm = get_community_by_name_or_id(community)
     if not comm:
-        raise click.BadParameter("There is no community by this name or ID: %s" 
-        % community)
-    if len(name)>255:
-        raise click.BadParameter("""NAME parameter is longer than the 255 
+        raise click.BadParameter("There is no community by this name or ID: %s"
+                                 % community)
```
emanueldima: I agree; it's just ocd from my part, trying to satisfy the linter.

b2share/modules/schemas/api.py
```diff
@@ -530,36 +530,36 @@ def create_version(cls, community_id, community_schema,
...
+
+                new_version = last_schema.version + 1
+                if root_schema_version is None:
+                    root_schema_version = last_schema.root_schema
+            except NoResultFound:
+                if root_schema_version is None:
+                    # there is no schema yet, the community is new.
+                    # Use the last RootSchema.
+                    try:
+                        root_schema_version = RootSchemaVersion.query.order_by(
```
emanueldima: This block is also changed in your recent PR, I think, so it will get another rewrite.

b2share/modules/schemas/cli.py
nharraud: Ok I missread. I thought it was a call to `helpers.validate_json_schema`. It might be better to move this call to the API. For now it works because only the CLI can create schemas but this won't always be the case.

emanueldima: yes, https://github.com/EUDAT-B2SHARE/b2share/issues/1392

b2share/modules/schemas/api.py
```diff
@@ -550,35 +550,38 @@ def create_version(cls, community_id, community_schema,
...
-                    community=community_id,
-                    root_schema=root_schema_version,
-                    community_schema=json.dumps(community_schema,
-                                                # avoid default whitespaces
-                                                separators=(',', ':')),
-                    version=new_version)
-                db.session.add(model)
-        except NoResultFound as e:
-            raise CommunitySchemaDoesNotExistError(id) from e
+        with db.session.begin_nested():
```
emanueldima: essentially this change only moves the try/except block close to the db query and fixes the raised error

emanueldima:
I improved the test by publishing a record for the newly created community with its own metadata schema.

PR #1387, opened 2017-02-21
schemas: backward compatibility check of schemas

### LINKED ISSUE #1076: api: check schemas backward compatibility using inveniosoftware/doschema.
We need to use the validation function in inveniosoftware/doschema for the validation of our block and community schemas.

The validation should be done in the programmatic API. The REST API should have an error handler which forwards the exception as a REST Exception [as it is done here](https://github.com/EUDAT-B2SHARE/b2share/blob/def0d0cce6d0677bab5c4354c44a7b81f478e467/b2share/modules/records/errors.py#L49).

Note: DoSchema is not released yet so the travis tests will work only in devel mode:
- Add `-e git+git://github.com/inveniosoftware/doschema.git#egg=doschema` in this file https://github.com/EUDAT-B2SHARE/b2share/blob/evolution/requirements-devel.txt
- Add doschema to setup.py.

The corresponding PR will of course be merged only once doschema is released so that the branch is not broken.

nharraud:
* NEW Checks backward compatibility of JSON Schemas. (based on #1096)
  (closes #1076)

Co-authored-by: PaulinaLach <paulina.malgorzata.lach@cern.ch>

### DIFF
--- a/b2share/modules/schemas/errors.py
+++ b/b2share/modules/schemas/errors.py
@@ -25,11 +25,18 @@
 
 from __future__ import absolute_import
 
+from doschema.errors import JSONSchemaCompatibilityError
+from invenio_rest.errors import FieldError, RESTException
 
-class B2ShareSchemasError(Exception):
+
+class B2ShareSchemasError(RESTException):
     """Base class for the B2Share Schemas module."""
-    pass
 
+    def __init__(self, description=None):
+        """Constructor"""
+        super(B2ShareSchemasError, self).__init__()
+        if description is not None:
+            self.description = description
 
 #
 # JSON SCHEMA ERRORS
@@ -37,8 +44,7 @@ class B2ShareSchemasError(Exception):
 
 class InvalidJSONSchemaError(B2ShareSchemasError):
     """Exception raised when a JSON Schema is considered as invalid."""
-    pass
-
+    code = 400
 
 #
 # ROOT SCHEMA ERRORS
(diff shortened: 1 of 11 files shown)


b2share/modules/schemas/errors.py
```diff
@@ -25,38 +25,45 @@
...
-class B2ShareSchemasError(Exception):
+
+class B2ShareSchemasError(RESTException):
     """Base class for the B2Share Schemas module."""
-    pass
 
+    def __init__(self, description=None, code=None):
+        """Constructor"""
+        super(B2ShareSchemasError, self).__init__()
+        self.code = code if code is not None else 500
```
emanueldima: This sets the default `code` value to 500 in all the subclasses' instances, which I believe is not the intended behaviour. Do you mean:

````
class B2ShareSchemasError(RESTException):
    code = 500
    def __init__(self, description=None, code=None):
        ....
        if code:
            self.code = code
````

File: b2share/modules/schemas/errors.py [hunk 2]
```diff
@@ -25,38 +25,45 @@
...
 
 #
 # JSON SCHEMA ERRORS
 #
 
 class InvalidJSONSchemaError(B2ShareSchemasError):
     """Exception raised when a JSON Schema is considered as invalid."""
-    pass
-
+    code = 400
```
emanueldima: Related to previous comment.

b2share/modules/schemas/errors.py
nharraud: Thanks for catching that. I'll improve it.

nharraud:
I will wait for PR 1384 to be merged before and rerun the tests, just in case.

PR #1395, opened 2017-02-24
Improve installation, augment demo cfg file, docs

emanueldima:


INSTALL.rst
```diff
@@ -33,43 +33,108 @@ account:
...
+.. code-block:: console
+
+    $ export B2SHARE_SECRET_KEY=mfTpGpQBwTTOmWCoZrYSuiJ1UzXq1d2L
+
+
+1.3 Prerequisite: configuring B2SHARE
 -------------------------------------
 
+B2SHARE needs connection to external services (e.g. elasticsearch, redis,
+nginx) to provide its features, and uses Docker for simplifying the creation
```
nharraud: we can list all the services here:
redis, elasticsearch, a database (postgresql or mysql), rabbitmq.
Nginx is not required by b2share. We just use it in order to provide https.

INSTALL.rst
```diff
@@ -33,43 +33,108 @@ account:
...
+section). If you are only trying to run a B2SHARE demo and you are certain
+that the data put in your B2SHARE installation doesn't need to be preserved,
+you can erase the ``services > b2share > volumes`` section from the
+``docker-compose.yml`` file.
+
 B2SHARE has a long list of configuration parameters, with default values
-available for inspection in the `b2share/config.py` file.
+available for inspection in the `b2share/config.py` file. By default, on the
+first run, the important parameters (that a site administrator is expected to
+modify) are copied to a separate ``/b2share-data/b2share.cfg``
```
nharraud: `/b2share-data/b2share.cfg` is inside the container.

demo/b2share_demo/config.py
```diff
@@ -21,42 +21,82 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-"""Demonstration configuration."""
+"""Demonstration configuration, to be changed by the site administrator"""
```
nharraud: Does it mean that the demo can be used production? This module was intended to be used only for demonstration purpose. Normally one shouldn't try to modify the config.py files. The two ways for users to modify the config are:
- the <instance_directory>/b2share.cfg file.
- environment variables.

PR #1398, opened 2017-02-24
views: flask-security endpoints disabled

### LINKED ISSUE #1397: views: do not register flask_security's endpoints
**Problem:**
Flask_security registers many endpoints which are not used in B2Share. Those endpoints are:
```
security.change_password                           HEAD,GET,POST,OPTIONS /api/accounts/settings/password/
security.confirm_email                             HEAD,GET,POST,OPTIONS /api/confirm/[token]
security.forgot_password                           HEAD,GET,POST,OPTIONS /api/lost-password/
security.login                                     HEAD,GET,POST,OPTIONS /api/login/
security.logout                                    HEAD,GET,OPTIONS     /api/logout/
security.register                                  HEAD,GET,POST,OPTIONS /api/signup/
security.reset_password                            HEAD,GET,POST,OPTIONS /api/lost-password/[token]/
```

Only the `logout` endpoint is used by the UI. Thus we should only add this endpoint.

**Solution:**
We can use [the `ACCOUNTS_REGISTER_BLUEPRINT` config variable provided by invenio_accounts](https://github.com/inveniosoftware/invenio-accounts/blob/b3efa4b6f3576d55ed60ead64736eab09a43e459/invenio_accounts/ext.py#L145).

It's better to do it before the V2.0 release as the additional endpoints could add some security issues.

nharraud:
* Disables flask-security endpoints as most of them were not used.
  Moves the logout endpoint to '/api/user/logout/'. (closes #1397)

### DIFF
--- a/b2share/config.py
+++ b/b2share/config.py
@@ -186,8 +186,8 @@ OAUTHCLIENT_REMOTE_APPS = dict(
     b2access=b2access.REMOTE_APP,
 )
 
-# Let Invenio Accounts register Flask Security
-ACCOUNTS_REGISTER_BLUEPRINT = True
+# Don't let Invenio Accounts register Flask Security
+ACCOUNTS_REGISTER_BLUEPRINT = False
 
 
 ACCOUNTS_REST_ASSIGN_ROLE_PERMISSION_FACTORY = \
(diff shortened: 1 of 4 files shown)


nharraud:
I tested manually the UI after the change. The logout still works.

Issue #1419, opened 2017-03-01
deployment: default files location not created automatically


nharraud:
**Problem:**
When the docker deployment is created with `LOAD_DEMO_COMMUNITIES_AND_RECORDS=0` or unset, there is no file location created. The result is that it is not possible to create buckets, which means that it is not possible to create draft records.

**Temporary solution:**
Once the deployment is started, start a shell on the b2share docker container.
```sh
$ docker exec -it dockerize_b2share_1 /bin/bash
```
Then create the location
```
b2share files add-location 'local' file:///usr/var/b2share-instance/files --default
```

emanueldima:
@nharraud Solved? or do you want to have it fixed in code?

nharraud:
Yes this is fixed for 2.0.1 but I don't think #1420 is fixed as people using v2.0.0 will still have the issue.

PR #1421, opened 2017-03-01
Fix missing location when demo not loaded

emanueldima:
NOT TESTED: this is a suggestion for fixing https://github.com/EUDAT-B2SHARE/b2share/issues/1419, let me know if you'd be happy with it.

### DIFF
--- a/dockerize/b2share.sh
+++ b/dockerize/b2share.sh
@@ -21,6 +21,9 @@ if [ ! -f /usr/var/b2share-instance/provisioned ]; then
 
     if [ "${LOAD_DEMO_COMMUNITIES_AND_RECORDS}" = "1" ]; then
         /usr/bin/b2share demo load_data
+    elif [ "${INIT_DB_AND_INDEX}" = "1" ]; then
+        # add default location
+        /usr/bin/b2share files add-location 'local' file:///usr/var/b2share-instance/files --default
     fi
 
     touch /usr/var/b2share-instance/provisioned


nharraud:
Before doing that I think we could search for the reason why the location is created for the demo and not without it. Just to understand and be sure that we are not adding another bug.

thijscobben:
And/or at least rename the variable "INIT_DB_AND_INDEX_AND_CREATE_EMPTY_LOCATION" or better maybe "INITIALIZE_B2SHARE" and document the things it does.
I would say, we discriminate to an "Empty setup" without load_demo that should work. Apparantly now it doesn't work because the location is not set. Then it makes sense to make two settings/commands: 1. Initialize empty but working system 2. Load demo data

nharraud:
@thijscobben This is what we were doing before (2 setups, one demo one not demo). We should be careful with it because there are at least 4 possible states:
- first start demo
- first start not demo
- restart demo
- restart not demo

We will have to teach our users the commands they have to run. Otherwise every time something goes wrong (ex: not enough disk space so the docker container crashes), the container will be in an unknown state and users will spam us and ask to fix things. IMHO it's better to have a clear documentation which explains every command. We can direct them to the Invenio documentation which is improving fast.

emanueldima:
TODO: investigate where or how the demo creates this default location

nharraud:
@thijscobben found the reason. The CLI command [`b2share demo load_data` adds a location](https://github.com/EUDAT-B2SHARE/b2share/blob/ae98bca812522453a1e5f4e708a881781bec845f/demo/b2share_demo/cli.py#L58).

emanueldima:
How do we solve this? I thing this PR is good enough.

nharraud:
@emanueldima Yes it looks good to me. Did you test it? I am currently rebuilding docker in order to do it.

emanueldima:
No, this is not tested yet, so it should not yet be merged, but it's good that we know how to move on.

Issue #1423, opened 2017-03-02
deployment: docker is bypassing iptables firewalls


bne86:
Hi,
due to the fact that docker is bypassing a linux firewall (on the host that we deploy b2share to we are dropping all connections that we do not explicitly allow), elasticsearch, rabbitmq, etc. are available via network. Due to the fact that we use some default credentials at some places, we should not expose the ports via docker-compose and instead only use the internal link mechanism.
There were some examples in the past where unprotected elasticsearch/mongodb servers were attacked, we should not make similar mistakes.

emanueldima:
already fixed? see https://github.com/EUDAT-B2SHARE/b2share/issues/1399

bne86:
No, you removed the explicit port mapping. Then he randomly uses other ports and those ports are still available to connect to.

emanueldima:
more info: https://github.com/moby/moby/issues/22054
bne86's solution is correct

emanueldima:
should be solved by https://github.com/EUDAT-B2SHARE/b2share/pull/1462, which includes @bne86's fix

PR #1424, opened 2017-03-02
rabbitmq and port hardening

bne86:
This PR addresses #1423 but also introduces the configuration of rabbitmq username/password. And, it should be possible to use alpine images for rabbitmq/redis.

### DIFF
--- a/dockerize/docker-compose.yml
+++ b/dockerize/docker-compose.yml
@@ -28,7 +28,7 @@ services:
             - "B2SHARE_CACHE_REDIS_HOST='redis'"
             - "B2SHARE_CACHE_REDIS_URL='redis://redis:6379/0'"
             - "B2SHARE_ACCOUNTS_SESSION_REDIS_URL='redis://redis:6379/1'"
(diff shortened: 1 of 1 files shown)


emanueldima:
Why make all these drastic changes? Isn't it better to keep the old config and use ports without publishing them to the host? i.e. use expose? https://docs.docker.com/compose/compose-file/compose-file-v2/#/expose

bne86:
The compose file that i started with was vulnerable. Both because a default username/password was used and because the port was accessible through the (maybe only local) network.

In our environment completely removing the port section (simply relying on the links mechanism) works well, using "expose" was not necessary. So, i did not see any reason for this.

Simply removing the ports section was the fastest but tested solution for the given problem, i also do not think that is it "drastic". But, feel free to use "expose" instead. I am not a b2share developer, i only want to share the solution that works for our deployment.

emanueldima:
Indeed, thank you for the PR and for signalling the issue; we have to discuss internally how to fix this best.

emanueldima:
merged via https://github.com/EUDAT-B2SHARE/b2share/pull/1462

PR #1429, opened 2017-03-07
post-release bump and AUTHORS.rst fix

nharraud:
post-release bump and AUTHORS.rst fix

### DIFF
--- a/b2share/version.py
+++ b/b2share/version.py
@@ -21,4 +21,4 @@
 """Version number."""
 
 
-__version__ = "2.0.0"
+__version__ = "2.0.1.dev20170307"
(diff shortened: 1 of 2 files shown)


b2share/version.py
```diff
@@ -21,4 +21,4 @@
 """Version number."""
 
 
-__version__ = "2.0.0"
+__version__ = "2.0.1.dev20170307"
```
emanueldima: Should this follow http://semver.org/#spec-item-9? (I don't care too much tbh).

nharraud: This is the `.dev` part [as described here](https://www.python.org/dev/peps/pep-0440/#public-version-identifiers)

PR #1436, opened 2017-03-16
Bugfix in the set_schema cli command

thijscobben:
set_schema worked for communities that had a schema and it worked for newly created communities but it didn't work for the EUDAT community that did not really have a schema but community_schema refered to nothing with an empty properties element.

so, after demo data loading, set_schema EUDAT <json> works.


emanueldima:
Manually rebased and merged, see https://github.com/EUDAT-B2SHARE/b2share/commit/975681bed802379931e95dcdc28209c0c4cf286b

PR #1445, opened 2017-03-28
Improve record cli command to handle unregistered dois

emanueldima:


tests/b2share_functional_tests/test_record_cli.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+        runner = CliRunner()
+        script_info = ScriptInfo(create_app=lambda info: app)
+
+        # Run 'check_record_doi' command for all records
+        result = runner.invoke(check_dois, ['-a'], obj=script_info)
+        if result.exit_code != 0:
+            print(result.output)
+        assert result.exit_code == 0
```
nharraud: This does not test if the command succeeds, just that it does not return an error code.

tests/b2share_functional_tests/test_record_cli.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # Run 'check_record_doi' command for only one record
+        result = runner.invoke(check_dois, ['-r', record_id_list[0]],
+                               obj=script_info)
+        if result.exit_code != 0:
+            print(result.output)
+        assert result.exit_code == 0
```
nharraud: This does not test if the command succeeds, just that it does not return an error code.

tests/b2share_functional_tests/test_schema_cli.py
```diff
@@ -63,7 +63,7 @@
                   ({'config': {'PREFERRED_URL_SCHEME': 'http'}})]
 
 @pytest.mark.parametrize('app', configurations, indirect=['app'])
-def _test_existing_community_set_schema_cmd(app, test_communities):
+def test_existing_community_set_schema_cmd(app, test_communities):
```
nharraud: Could this be in a separate commit? It will make finding possible bugs easier.

b2share/modules/records/cli.py
```diff
@@ -48,22 +50,31 @@ def b2records():
...
+@click.option('-u', '--update', is_flag=True, default=False)
+def check_dois(record, allrecords, update):
+    """ Checks that DOIs of records in the current instance are registered.
+    """
+    if record:
+        record = Record.get_record(record)
+        check_record_doi(record, update)
+    elif allrecords:
+        click.secho('checking DOI for all records')
+        for record in list_db_published_records():
```
nharraud: Does this work for multiple records? There is a commit after each operation.

tests/b2share_functional_tests/test_record_cli.py
```diff
@@ -0,0 +1,94 @@
...
+           ('Accept', 'application/json')]
+patch_headers = [('Content-Type', 'application/json-patch+json'),
+                 ('Accept', 'application/json')]
+
+def test_record_doi_cli(app, test_communities, login_user):
+    """Test adding a community and setting its schema using CLI commands."""
+    with app.app_context():
+        community_name = test_communities.popitem()[0]
+        community = Community.get(name=community_name)
+        com_admin = create_user('com_admin', roles=[community.admin_role])
```
nharraud: You can use [`test_community`](https://github.com/EUDAT-B2SHARE/b2share/blob/4d988f7a1b5d8678c234553f23685cc70b001363/tests/conftest.py#L205) fixture which creates the com_admin already.

tests/b2share_functional_tests/test_record_cli.py
```diff
@@ -0,0 +1,94 @@
...
+                assert draft_submit_res.status_code == 200
+                draft_publish_res = client.patch(
+                    draft_url,
+                    headers=patch_headers,
+                    data=json.dumps([{
+                        "op": "replace", "path": "/publication_state",
+                        "value": PublicationStates.published.name
+                    }]))
+                assert draft_publish_res.status_code == 200
+                record_id_list.append(record_id)
```
nharraud: You can use [`test_records`](https://github.com/EUDAT-B2SHARE/b2share/blob/4d988f7a1b5d8678c234553f23685cc70b001363/tests/conftest.py#L400) which already creates some published records.

tests/b2share_functional_tests/test_record_cli.py
emanueldima: I agree, but I don't see how to test the actual DOI allocation. This test just makes sure that the iteration through the records works.

tests/b2share_functional_tests/test_record_cli.py
emanueldima: same as above

tests/b2share_functional_tests/test_schema_cli.py
emanueldima: agree

tests/b2share_functional_tests/test_record_cli.py
nharraud: It should be possible to mock some functions and check that they have been called. The CLI part might make things a little more complex.

b2share/modules/records/cli.py
emanueldima: I don't know, why would it not work? The tests pass even when it's called.

tests/b2share_functional_tests/test_record_cli.py
```diff
@@ -0,0 +1,90 @@
...
+
+        def _my_ref(doi_value):
+            nonlocal ref_called
+            ref_called += 1
+
+        def _my_register(doi, url, doc):
+            nonlocal register_called
+            register_called += 1
+
+        records_cli._datacite_doi_reference = _my_ref
```
nharraud: This can be done in a better way with `unittest.mock`.
```python
with mock.patch.multiple('records_cli', reference='_datacite_doi_reference', register='_datacite_register_doi') as mocks:
    _datacite_register_doi = mocks['register']
    _datacite_doi_reference = mocks['reference']
    _datacite_doi_reference = ...
    ...
    assert datacite_doi_reference.assert_called_with(<doi_value>)
    assert datacite_doi_reference.assert_called_with(<doi, url, doc>)
```
This enable to replace the functions only for this test even if the test fails and raise an exception.

emanueldima:
@nharraud Please re-review

PR #1452, opened 2017-04-12
improvement to rabbitmq configuration

thijscobben:
The entrypoints prevent a script dockerize-entrypoint.sh from Dockerhub / rabbitmq to run, this is necessarily to prevent (new) default rabbitmq behaviour to only allow connections for guest user from localhost. Moreover, added settings so administrators can set different values for default username and password than guest / guest which is a security improvement.


emanueldima:
merged via https://github.com/EUDAT-B2SHARE/b2share/pull/1462

PR #1470, opened 2017-05-05
Switch to master branch in docs and scripts

emanueldima:


nharraud:
**NOTE FOR SUPPORT**: Also note that the users who try to update using git might get confused as the evolution branch is not the one they should use anymore. But it is not a problem as they shouldn't upgrade using git command but cloning a new version as explained in INSTALL.rst. We should answer this when answering support requests concerning this problem.

emanueldima:
I don't think it's good to change the virtualenv name because then our environment would have to be recreated. I certainly don't want to do that.

nharraud:
@emanueldima it's a one time operation. Of course we can live with it, but the creation of the virtualenv is anyway a part which prevents me from using those scripts. I have at least 3 or 4 virtualenv just for B2SHARE. I created one just today in order to test new invenio modules without destroying my stable virtualenv.

I often create a new virtualenv when I want to test a PR without having installed python modules interfering with it (I often use experimental invenio modules when I integrate them).

emanueldima:
I polled the dev team on our meeting and most people think it's better to rename the virtualenv to just `b2share`, so I'll do that.

PR #1485, opened 2017-05-29
split dockerize folder: build image + run

emanueldima:
The current `dockerize` folder is changed to only keep files related to building of the b2share docker image. All the other files, related to running a demo installation of b2share, are moved into the `dockerize-repository` folder.

In a subsequent PR, the `dockerize-repository` folder will be moved to the https://github.com/EUDAT-B2SHARE/dockerize repository and the https://github.com/EUDAT-B2SHARE/package-tests repository will be updated to use it for testing.


emanueldima:
@nharraud review pls?

INSTALL.rst
```diff
@@ -7,14 +7,15 @@ B2Share installation
 ===============================================
 
 
-1.0. Prerequisite: clone B2SHARE
---------------------------------
+1.0. Prerequisite: clone `dockerize`
+------------------------------------
 
-Current installation process requires a local clone of B2SHARE.
+The current installation process requires a local clone of the `dockerize`
```
nharraud: The docker image can be used separately. Thus I would rather write something like this:

===

The b2share docker image can be used as is and is available on dockerhub as [`eudatb2share/b2share`](https://hub.docker.com/r/eudatb2share/b2share/). However it requires many different services like elasticsearch, postgresql, redis and rabbitmq, thus we provide a standard deployment procedure which creates a demonstration environment.
The next steps will describe how to use this deployment.

First clone the `dockerize` repository:

===

dockerize-repository/docker-compose.yml
```diff
@@ -12,9 +12,7 @@ services:
             - "5432"
 
     b2share:
-        build:
-            context: ..
-            dockerfile: dockerize/Dockerfile
+        image: eudatb2share/b2share:2.0.1
```
nharraud: I am a little at loss here. If the deployment procedure is in the dockerize repository why do we need a docker-compose which also deploy production environment?

The only docker-compose which seems useful to me here is one for development, thus it should use local code. Otherwise we will be in the same situation as before the last docker scripts refactoring when we could not test our modifications before merging PRs. In fact it would be even worse here as we would have to release on docker-hub before being able to test.

nharraud: I see that EUDAT-B2SHARE/dockerize is empty. Maybe this docker-compose file is intended to be there?

If yes, I would still keep one docker-compose file either in this repo or the other just for testing local code in docker.

emanueldima: The `dockerize-repository` is what the users will clone and use as demo. The contents of this
folder will go as it is into the `EUDAT-B2SHARE/dockerize` repository (see first comment). For local testing we'll use `devenv`.

Issue #1477, opened 2017-05-10
Non-understandable error message when community metadata is missing


emanueldima:
Reported by Europeana. The returned HTTP message is a validation error (missing field xxx) which is not at all clear:

>`{
   "message": "Validation error.",
   "errors": [
    {
      "message": "'2a01ee91-36fe-4edb-9734-73d22ac78821' is a required property", 
       "field": "2a01ee91-36fe-4edb-9734-73d22ac78821"
    }
                 ], 
  "status": 400
}
`

emanueldima:
https://github.com/EUDAT-B2SHARE/b2share/pull/1486

PR #1486, opened 2017-05-30
Improve some error messages when publishing records

emanueldima:


emanueldima:
fix for https://github.com/EUDAT-B2SHARE/b2share/issues/1477

b2share/modules/deposit/api.py
```diff
@@ -224,6 +226,25 @@ def submit(self, commit=True):
         if commit:
             self.commit()
 
+    def validate_before_publishing(self):
+        """Common sense validations before submitting. These would be also
```
nharraud: There will probably be other messages considered cryptic. Can't we catch the exception raised by the jsonschema library and raise our own instead? It would avoid reimplementing every validation ourselves.

emanueldima: We already do that, the error users get now looks like this: 
````
{ "message": "Validation error.", "errors": [ { "message": "'2a01ee91-36fe-4edb-9734-73d22ac78821' is a required property", "field": "2a01ee91-36fe-4edb-9734-73d22ac78821" } ], "status": 400 }
````

When the field has some semantic meaning (e.g. `open_access`) the error is easy to understand, but for this special case of community specific metadata blocks specified by uuids, the error is confusing.

Comment (nharraud) [inline b2share/modules/deposit/api.py]: I know we replace some of the messages, what I meant is "can we do a special case for this error when we replace the error message"? Like using a regular expression on the error message to catch specifically the one we want to translate.

**Comment by emanueldima** [comment]
@nharraud redone based on your comments, please re-review

File: tests/b2share_unit_tests/deposit/test_deposit_rest.py [hunk 2]
```diff
@@ -177,6 +177,42 @@ def test_deposit_submit(app, test_records_data, draft_deposits, test_users,
...
+                        "op": "remove", "path": '/'+field,
+                        }, {
+                        "op": "replace", "path": "/publication_state",
+                        "value": PublicationStates.submitted.name
+                    }]),
+                    headers=headers)
+                assert draft_patch_res.status_code == 400
+                draft_patch_error = json.loads(
+                    draft_patch_res.get_data(as_text=True))
+                # import ipdb; ipdb.set_trace()
```
nharraud: remaining debug code

b2share/modules/records/errors.py
```diff
@@ -62,12 +63,28 @@ class AnonymousDepositSearch(B2ShareRecordsError):
...
             try:
                 field = err.message.split('\'')[1]
+                fieldpath = (fieldpath + '/' if fieldpath else '') + field
             except IndexError:
                 pass # ignore
+        if err.validator == 'required' and len(err.path) == 1 and \
+                err.path[0] == 'community_specific' and is_valid_uuid(field):
+            message = 'The "community_specific" metadata object must contain '\
+                      'an object named "{}" containing the '\
+                      'community-specific metadata fields'.format(field)
```
nharraud: It would be interesting to have a test checking that the message has been changed properly. Maybe I missed it.

nharraud:
@emanueldima Some minor changes but otherwise it looks good to me 👍

b2share/modules/records/errors.py
emanueldima: added it

b2share/modules/records/errors.py
```diff
@@ -62,12 +63,28 @@ class AnonymousDepositSearch(B2ShareRecordsError):
...
-            FieldError(field, err.message)
+            FieldError(fieldpath, message)
         ])
+
+
+def is_valid_uuid(string):
+    try:
+        uuid.UUID(string)
+        return True
+    except:
```
nharraud: One thing I didn't see before. Can you catch `ValueError` here instead of an empty except?

emanueldima: `uuid.UUID(0)` raises AttributeError...

nharraud: then `ValueError` and `AttributeError`. The error is `ValueError` is raised when the string is not a valid UUID.

nharraud: `except` should always be used with an exception, at the very least `Exception`, otherwise you would also catch important exceptions like `KeyboardInterrupt` and `SystemExit`.

emanueldima:
@nharraud Should be now fixed

Issue #1432, opened 2017-03-09
UI: missing title (just an empty space: ' ')


emanueldima:
It's possible to create in the UI a title containing non-printing chars, and the list of recent records looks ugly without a visible title.

PR #1492, opened 2017-06-21
Don't allow empty spaces for record title/fields

SarahBA:
fixed https://github.com/EUDAT-B2SHARE/b2share/issues/1432

### DIFF
--- a/webui/src/components/editrecord.jsx
+++ b/webui/src/components/editrecord.jsx
@@ -168,6 +168,10 @@ const EditRecord = React.createClass({
             }
         }
         console.assert(!Array.isArray(value));
+        if(typeof value === 'string' || value instanceof String) {
+            value = value.replace(/^\s+/, '').replace(/\s+$/, ' ') ;
+        }
+        
         r = value !== undefined ? r.setIn(path, value) : r.deleteIn(path);
         const errors = this.state.errors;
         const pathstr = path.join('/');
--- a/webui/src/components/newrecord.jsx
+++ b/webui/src/components/newrecord.jsx
@@ -24,7 +24,7 @@ export const NewRecordRoute = React.createClass({
 
     createAndGoToRecord(event) {
         event.preventDefault();
-        if (!this.state.title.length) {
+        if (!this.state.title.length || !this.state.title.trim()) {
             this.setError('title', "Please add a (temporary) record title");
             return;
         }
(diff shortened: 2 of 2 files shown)


webui/src/components/editrecord.jsx
```diff
@@ -168,7 +168,14 @@ const EditRecord = React.createClass({
             }
         }
         console.assert(!Array.isArray(value));
-        r = value !== undefined ? r.setIn(path, value) : r.deleteIn(path);
+        let trimmedValue;
+        if(typeof value === 'string' || value instanceof String) {
+            trimmedValue = value.replace(/^\s+/, '').replace(/\s+$/, ' ') ;
```
emanueldima: let's use trim() for this (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/trim)

webui/src/components/newrecord.jsx
```diff
@@ -24,7 +24,7 @@ export const NewRecordRoute = React.createClass({
 
     createAndGoToRecord(event) {
         event.preventDefault();
-        if (!this.state.title.length) {
+        if (!this.state.title.length || !this.state.title.replace(/^\s+/, '').replace(/\s+$/, '')) {
```
emanueldima: trim()

webui/src/components/newrecord.jsx
```diff
@@ -35,7 +35,7 @@ export const NewRecordRoute = React.createClass({
         const json_record = {
             community: this.state.community_id,
             titles: [{
-                title: this.state.title,
+                title: this.state.title.replace(/^\s+/, '').replace(/\s+$/, ''),
```
emanueldima: trim()

webui/src/components/editrecord.jsx
SarahBA: Using trim() here doesn't allow to have multiple words because the state updates after pressing any key and if it is a space, it will be removed by trim(). Then user need to write them like a compound word and then move the cursor back and add a space which I think it's annoying

emanueldima: Ok with this one.

webui/src/components/editrecord.jsx
```diff
@@ -168,7 +168,14 @@ const EditRecord = React.createClass({
             }
         }
         console.assert(!Array.isArray(value));
-        r = value !== undefined ? r.setIn(path, value) : r.deleteIn(path);
+        let trimmedValue;
```
emanueldima: Please just change the default `value` here and not add this `trimmedValue` variable. It's less code and less buggy.

SarahBA: fixed

Issue #1363, opened 2017-02-07
User not able to delete API token


thijscobben:
This can compromise security if API token trust is breached. Therefore I label it as urgent.

emanueldima:
It's possible to delete it using the HTTP API (DELETE /api/users/current/tokens/$TOKEN)

PR #1516, opened 2017-07-13
Delete a user token

SarahBA:
closing https://github.com/EUDAT-B2SHARE/b2share/issues/1363

### DIFF
--- a/webui/src/components/user.jsx
+++ b/webui/src/components/user.jsx
@@ -2,6 +2,7 @@ import React from 'react/lib/ReactWithAddons';
 import { Link } from 'react-router'
 import { serverCache, Error, loginURL, notifications } from '../data/server';
 import { CommunityAdmin } from './community_admin.jsx'
+import { Wait, Err } from './waiting.jsx';
 
 const PT = React.PropTypes;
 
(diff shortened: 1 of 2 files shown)


webui/src/components/user.jsx
```diff
@@ -156,46 +157,47 @@ export const UserProfile = React.createClass({
...
+            serverCache.newUserToken(tokenName, token => {
+                notifications.success("A new Token is created");
+                this.setState({token});
             });
         }
     },
 
     renderToken(t) {
-        return <li className="list-group-item" key={t.id}>{t.name}</li>;
+        return <div key={t.id}> <Token token_id={t.id} token_name={t.name} removeToken={this.removeToken} /> </div>
```
emanueldima: I think we can we remove the topmost div.

webui/src/components/user.jsx
```diff
@@ -156,46 +157,47 @@ export const UserProfile = React.createClass({
...
+        if (!tokenName) {
             notifications.danger('Please choose a name for the new token');
         } else {
             notifications.clearAll();
-            serverCache.newUserToken(this.state.name, token => {
-                const tokenList = this.state.tokenList;
-                tokenList.push(token);
-                this.setState({token, tokenList})
+            serverCache.newUserToken(tokenName, token => {
+                notifications.success("A new Token is created");
```
SarahBA: Right, it's not necessary.

webui/src/components/user.jsx
SarahBA: true, updating the PR right now

SarahBA: No wait, although they get a key in 'Token' render method, `key={this.props.token_id}`, each of the elements in the iterator still needs a key in `TokenList`component. That's why I added the div there.

emanueldima: Ok.

PR #1521, opened 2017-07-25
Improve language select

SarahBA:


SarahBA:
I marked it as "in progress" because right now user needs to click 2 times on the dropdown menue to be able to select a language from the full list which I'm not happy with. Maybe it's better to use something like react-select-2  @emanueldima What do you think?

webui/src/components/editrecord.jsx
```diff
@@ -291,13 +291,20 @@ const EditRecord = React.createClass({
             field = this.renderEmbargoField(schema, path);
         } else if (objEquals(path, ['language']) || objEquals(path, ['language_code'])) {
             const languages = serverCache.getLanguages();
+            const langs = {"languages": [["arb", "Standard Arabic"], ["ben", "Bengali"], ["bul", "Bulgarian"], ["ces", "Czech"], ["cmn", "Mandarin Chinese"], ["dan", "Danish"], ["deu", "German"],
```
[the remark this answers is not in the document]
SarahBA: These are the world’s most common languages (like top 20) + the language of countries that some of our communities are from.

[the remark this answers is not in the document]
emanueldima: Good, can you put a comment in the code which says that? It would be useful for a future reader of this code.

webui/src/components/selectbig.jsx
```diff
@@ -13,25 +13,46 @@ export const SelectBig = React.createClass({
...
+            dataList: [],
+            extend: false,
+            open: false,
         };
     },
 
-    select: function (val) {
-        this.props.onSelect(val.id);
+    select(val) {
+        console.log("val = ", val)
```
emanueldima: to remove

webui/src/data/server.js
```diff
@@ -652,6 +652,9 @@ class ServerCache {
         if (!disciplines) {
             this.getters.disciplines.autofetch();
         }
+        if(disciplines && disciplines.filter((v) => { if(v.id === "-") { return v.id } }).length == 0 ){
```
emanueldima: I think `(v) => { if(v.id === "-") { return v.id } }` can be rewritten as `v => v.id === "-"` (will return true if the id is '-', false otherwise).

webui/src/components/editrecord.jsx
```diff
@@ -291,14 +291,28 @@ const EditRecord = React.createClass({
...
+                                            ["pol", "Polish"], ["por", "Portuguese"], ["ron", "Romanian"], ["rus", "Russian"], ["slk", "Slovak"], ["slv", "Slovenian"], ["spa", "Spanish"],
+                                            ["swe", "Swedish"], ["tam", "Tamil"], ["tel", "Telugu"], ["tur", "Turkish"], ["urd", "Urdu"], ["vie", "Vietnamese"], ["...", "Load Full List"]]};
+            const common_langs = langs.languages.map(([id, name]) => ({id, name}));
             field = (languages instanceof Error) ? <Err err={languages}/> :
-                <SelectBig data={languages}
-                    onSelect={x=>this.setValue(schema, path, x)} value={this.getValue(path)} />;
+                <SelectBig data={common_langs}
+                            list_extend={languages}
+                            extendable={true}
+                            onSelect={ (x) => { x ? this.setValue(schema, path, x) : this.setValue(schema, path, undefined) }}
```
emanueldima: compress the inner closure: `x => this.setValue(schema, path, x ? x : undefined)`

webui/src/components/editrecord.jsx
```diff
@@ -291,14 +291,28 @@ const EditRecord = React.createClass({
...
+                            value={this.getValue(path)} />;
         } else if (path.length === 2 && path[0] === 'disciplines') {
             const disciplines = serverCache.getDisciplines();
             field = (disciplines instanceof Error) ? <Err err={disciplines}/> :
                 <SelectBig data={disciplines}
-                    onSelect={x=>this.setValue(schema, path, x)} value={this.getValue(path)} />;
+                            list_extend={null}
+                            extendable={false}
+                            onSelect={ (x) => { x ? this.setValue(schema, path, x) : this.setValue(schema, path, undefined) }}
```
emanueldima: compress it as above

hevp:
This will be implemented in a different way. This PR is therefore obsolete.

PR #1523, opened 2017-07-31
Add versioning support, tests and UI

emanueldima:


b2share/modules/deposit/api.py
```diff
@@ -112,6 +128,36 @@ def create(cls, data, id_=None):
...
+            raise InvalidDepositError(
+                'Field "_parent_pids" cannot be set.')
+        version_master, prev_version = None, None
+        if version_of:
+            version_master, prev_version = find_version_master_and_previous_record(version_of)
+
+        if prev_version:
+            if not get_version_parent_pid(prev_version):
+                # previous version needs updating with the versioning parent pid
+                set_version_parent_pid(prev_version, version_master.parent)
```
nharraud: We can instead migrate all records.

b2share/modules/deposit/api.py
```diff
@@ -246,3 +300,72 @@ def create_file_pids(record_metadata):
...
+
+    version_master = retrieve_version_master(child_pid)
+    if not version_master:
+        # the old record is NOT versioned yet
+        parent_pid = VersionMasterPidProvider.create().pid
+        parent_pid.redirect(child_pid)
+        version_master = PIDVersioning(parent=parent_pid)
+        version_master.insert_child(child=child_pid, index=-1)
+
+    prev_pid = version_master.last_child
```
nharraud: It doesn't seem to me that this code use the "draft" feature provided by invenio-pidrelations.

b2share/modules/deposit/api.py
```diff
@@ -112,6 +128,36 @@ def create(cls, data, id_=None):
...
+                raise InvalidDepositError(
+                    'Community ID must be identical with '
+                    'the previous version\'s community ID.')
+            if current_user.id not in prev_version['_deposit']['owners']:
+                raise OwnershipVersioningError()
+            patch = data
+            data = copy.deepcopy(prev_version.model.json)
+            # eliminate _deposit, _files, _oai, _pid, etc.
+            data = {k: v for k, v in data.items() if not k.startswith('_')}
+            data.update(patch)
```
nharraud: Do we really want to always make a patch. This might create issue with records created via the REST API.

nharraud: Do we really want to keep by default the values of the previous record? This will create problems with records created via the REST API.

emanueldima: This is a very important feature for usability, we should do it even if it costs us. But I don't yet see what problems it can create, can you clarify?

nharraud: If you create a record via REST API I expect that you already have all the data. Ex: LTER records. There might be unexpected side effects if a field is renamed in the community schema. The field with the old name would be copied without the client noticing.
For UI at least the user checks the content before publishing. Then copying makes more sense IMHO. Although there would still be the issue of schema change.

emanueldima: Yes you're right, this should not be handled here in the programmatic api but in the UI, I'll make the changes.

b2share/modules/deposit/api.py
```diff
@@ -112,6 +137,49 @@ def create(cls, data, id_=None):
...
+                        version_master.draft_child
+                    )
+                raise exc
+        else:
+            # create parent PID
+            parent_pid = RecordUUIDProvider.create().pid
+            version_master = PIDVersioning(parent=parent_pid)
+            version_master.insert_draft_child(child=rec_pid)
+            # Mark this deposit as the first version
+            # TODO(edima): remove 'first_version' from json?
```
emanueldima: TODO: remove 'first_version', provide the versions link all the time

b2share/modules/deposit/api.py
```diff
@@ -112,6 +137,39 @@ def create(cls, data, id_=None):
...
+        # Retrieve reserved record PID (The record PID value is the same
+        # as the one of the deposit)
+        rec_pid = RecordUUIDProvider.get(data['_deposit']['id']).pid
+        version_master, prev_version = None, None
+        if version_of:
+            version_master, prev_version = \
+                find_version_master_and_previous_record(version_of)
+            if current_user.id not in prev_version['_deposit']['owners']:
+                raise OwnershipVersioningError()
+            try:
```
emanueldima: We also need to check here that the `community` has not been changed.

webui/src/data/server.js
```diff
@@ -529,7 +554,7 @@ class ServerCache {
     }
 
     getRecord(id) {
-        this.getters.record.get(id).fetch();
+        this.getters.record.get(id).forceFetch();
```
emanueldima: Can we avoid this change? It would hurt the performance when re-displaying records.

webui/src/components/versions.jsx
```diff
@@ -0,0 +1,94 @@
...
+        return (
+            <div style={style}>
+                You are now editing a draft of
+                <Link onClick={e=>browser.gotoRecord(versions[0].id)}> this published record</Link>.
+            </div>
+        );
+    },
+});
+
+const PublishedVersions = React.createClass({
```
emanueldima: `return false` is sufficient here, we probably don't need to return an empty div.

webui/src/components/schema.jsx
```diff
@@ -6,7 +6,8 @@ import { Wait, Err } from './waiting.jsx';
 
 const except = {'$schema':true, 'community_specific':true, 'owner':true,
                 '_internal':true, '_deposit':true, '_files':true,
-                '_pid':true, '_oai':true, 'publication_state': true, };
+                '_pid':true, '_oai':true, 'publication_state': true,
+                '_parent_pids': true, };
```
emanueldima: This change is no longer necessary (`_parent_pids` was removed).

b2share/modules/schemas/root_schemas/root_schema_v0.json
```diff
@@ -228,6 +228,10 @@
             },
             "_files": {
                 "type": "array"
+            },
+            "_parent_pids": {
```
emanueldima: This change is no longer necessary (`_parent_pids` has been removed)

webui/src/components/versions.jsx
```diff
@@ -0,0 +1,102 @@
...
+        if(versions.length > 0){
+            return (
+                <div style={style}>
+                    You are now editing a draft of
+                    <Link onClick={e=>browser.gotoRecord(versions[0].id)}> this published record</Link>.
+                </div>
+            );
+        }
+        else{
+            return (
```
emanueldima: `return false` here

webui/src/data/server.js
dinosk: @emanueldima Right, I removed this. I added it in an attempt to get a fresh version of the record page because the cached one was outdated. Navigating to a directly older version rendered the page as when that was the latest one. The versions are shown correctly now but I can't get around this annoying rerender of the dropdown list and the create button. 
versioning_lag

b2share/modules/deposit/api.py
```diff
@@ -224,6 +305,26 @@ def submit(self, commit=True):
...
+        record_pid = RecordUUIDProvider.get(pid_value).pid
+        version_child = PIDVersioning(child=record_pid)
+        version_master = PIDVersioning(parent=version_child.parent)
+        if version_master is not None \
+                and version_master.draft_child.pid_value == pid_value:
+            pid = PersistentIdentifier.query.\
+                filter_by(pid_value=pid_value, pid_type='b2dep').one_or_none()
+            if not pid:
+                raise ValueError()
+            version_master.remove_draft_child()
```
nharraud: What happens if we delete an intermediate deposit?

b2share/modules/deposit/api.py
```diff
@@ -224,6 +305,26 @@ def submit(self, commit=True):
...
             self.commit()
 
+    @classmethod
+    def delete(self, pid_value):
+        """Delete a deposit."""
+        record_pid = RecordUUIDProvider.get(pid_value).pid
+        version_child = PIDVersioning(child=record_pid)
+        version_master = PIDVersioning(parent=version_child.parent)
+        if version_master is not None \
+                and version_master.draft_child.pid_value == pid_value:
```
nharraud: We should allow the deletion of intermediate deposits.
```python
if version_master is not None
```

b2share/modules/deposit/api.py
```diff
@@ -224,6 +305,26 @@ def submit(self, commit=True):
...
+                filter_by(pid_value=pid_value, pid_type='b2dep').one_or_none()
+            if not pid:
+                raise ValueError()
+            version_master.remove_draft_child()
+            pid.delete()
+            # hard delete the record pid
+            db.session.delete(record_pid)
+            if version_child.parent.status == PIDStatus.RESERVED:
+                db.session.delete(version_child.parent)
+            db.session.commit()
```
nharraud: deposit should also be removed from the ES index.

b2share/modules/deposit/api.py
```diff
@@ -246,3 +347,38 @@ def create_file_pids(record_metadata):
                 raise e
             else:
                 current_app.logger.warning(e)
+
+
+def find_version_master_and_previous_record(version_of):
+    """"""
+    try:
+        child_pid = RecordUUIDProvider.get(version_of).pid
```
nharraud: What happens `version_of` references a parent PID?

b2share/modules/deposit/api.py
```diff
@@ -112,6 +137,39 @@ def create(cls, data, id_=None):
...
             raise InvalidDepositError('"$schema" field should not be set.')
+
+        # Retrieve reserved record PID (The record PID value is the same
+        # as the one of the deposit)
+        rec_pid = RecordUUIDProvider.get(data['_deposit']['id']).pid
+        version_master, prev_version = None, None
+        if version_of:
+            version_master, prev_version = \
+                find_version_master_and_previous_record(version_of)
+            if current_user.id not in prev_version['_deposit']['owners']:
```
nharraud: This should be checked by the permission factory.

b2share/modules/deposit/api.py
```diff
@@ -246,3 +347,38 @@ def create_file_pids(record_metadata):
...
+        if child_pid.status == PIDStatus.DELETED:
+            raise RecordNotFoundVersioningError()
+    except PIDDoesNotExistError as e:
+        raise RecordNotFoundVersioningError() from e
+
+    version_master = PIDVersioning(child=child_pid)
+
+    prev_pid = version_master.last_child
+    assert prev_pid.pid_type == RecordUUIDProvider.pid_type
+    prev_version = Record.get_record(prev_pid.object_uuid)
```
nharraud: What happens version_of references a parent PID? I expect a 500 error.

b2share/modules/deposit/api.py
```diff
@@ -246,3 +347,38 @@ def create_file_pids(record_metadata):
...
+    version_master, prev_version = \
+        find_version_master_and_previous_record(version)
+    data = prev_version.model.json
+    # eliminate _deposit, _files, _oai, _pid, etc.
+    data = {k: v for k, v in data.items() if not k.startswith('_')}
+    del(data['publication_state'])
+    if 'publication_date' in data:
+        del(data['publication_date'])
+    del(data['$schema'])
+    return data
```
nharraud:
```python
return copy.deepcopy(data)
```

b2share/modules/deposit/api.py
```diff
@@ -246,3 +347,38 @@ def create_file_pids(record_metadata):
...
+    if prev_pid.pid_value != version_of:
+        raise IncorrectRecordVersioningError(prev_pid)
+    return version_master, prev_version
+
+
+def copy_data_from_previous(version):
+    """Copy metadata from previous record version."""
+    version_master, prev_version = \
+        find_version_master_and_previous_record(version)
+    data = prev_version.model.json
```
nharraud: data = copy.deepcopy(prev_version.model.json)

b2share/modules/deposit/errors.py
```diff
@@ -33,3 +35,59 @@ class B2ShareDepositError(RESTValidationError):
...
+
+
+class VersioningDepositError(B2ShareDepositError):
+    """Versioning related exception."""
+
+
+class OwnershipVersioningError(VersioningDepositError):
+    """Exception raised when ownership changed in a version chain.
+    Currently the ownership must be a constant."""
+    code = 403
```
nharraud: We need to be sure that this is not raised if a user is anonymous (not authenticated), as it would then be 401 and not 403.

b2share/modules/deposit/permissions.py
```diff
@@ -92,12 +92,20 @@ def update_deposit_metadata_need_factory(community, publication_state):
                                  publication_state=publication_state)
 
 
+def delete_deposit_need_factory(community, publication_state):
```
nharraud: Let's check if we need that. The permissions already existed before.

b2share/modules/deposit/permissions.py
```diff
@@ -319,7 +326,8 @@ class DeleteDepositPermission(DepositPermission):
     def _load_additional_permissions(self):
         permission = StrictDynamicPermission()
```
nharraud: not needed anymore.

b2share/modules/records/api.py
```diff
@@ -0,0 +1,128 @@
...
+    def delete(cls, pid, record):
+        """Delete a published record.
+
+        params:
+        pid: PersistentIdentifier
+        record: invenio_records_files.api.Record instance
+        """
+        # Fetch deposit id from record and resolve deposit record and pid.
+        depid = b2share_deposit_uuid_fetcher(None, record)
+        if not depid:
```
nharraud: The deposit could have been deleted before.

b2share/modules/records/api.py
```diff
@@ -0,0 +1,128 @@
...
+        except NotFoundError:
+            pass
+
+        # remove record from versioning chain
+        # if the parent has no other children and no draft child
+        # delete it from the db completely
+        version_child = PIDVersioning(child=pid)
+        version_master = PIDVersioning(parent=version_child.parent)
+        if version_master is not None:
+            version_master.remove_child(pid)
```
nharraud: We should keep the link between the parent and the child.
Use `PIDVersioning.update_redirect`.

b2share/modules/records/api.py
```diff
@@ -0,0 +1,128 @@
...
+            if not version_master.children.all() and \
+                    not version_master.draft_child:
+                db.session.delete(version_master.parent)
```
nharraud: The master PID should not be deleted. It will just redirect to a PID marked as DELETED and thus return 410.

b2share/modules/records/api.py
```diff
@@ -0,0 +1,128 @@
...
+            PersistentIdentifier.object_type == pid.object_type,
+            PersistentIdentifier.object_uuid == pid.object_uuid,
+        ).all()
+        for rec_pid in all_pids:
+            if not rec_pid.is_deleted():
+                rec_pid.delete()
+        depid.delete()
+
+        # Mark the record and deposit as deleted
+        record.delete()
```
nharraud: config's `B2SHARE_RECORDS_REST_ENDPOINTS.record_class` should be changed to use this record class. Then we should call `super` here.

b2share/modules/records/api.py
```diff
@@ -0,0 +1,128 @@
...
+        # Fetch deposit id from record and resolve deposit record and pid.
+        depid = b2share_deposit_uuid_fetcher(None, record)
+        if not depid:
+            abort(404)
+        depid, deposit = Resolver(
+            pid_type=depid.pid_type,
+            object_type='rec',
+            # Retrieve the deposit with the Record class on purpose as the current
+            # Deposit api prevents the deletion of published deposits.
+            getter=Record.get_record,
```
nharraud: Here it should be `Deposit.get_record` with Deposit being the B2SHARE Deposit api class.

b2share/modules/records/api.py
```diff
@@ -0,0 +1,128 @@
...
+                    id=version_master.last_child.object_uuid).one_or_none()
+                new_last_child_rec = Record(new_last_child_rec_meta.json,
+                                            model=new_last_child_rec_meta)
+                RecordIndexer().index(new_last_child_rec)
+
+        # Mark connected buckets as DELETED
+        if record.files:
+            record_bucket = record.files.bucket
+            deposit_bucket = deposit.files.bucket
+            Bucket.delete(deposit_bucket.id)
```
nharraud: We should keep the buckets for now just to be consistent with the Deposit.

b2share/modules/records/api.py
```diff
@@ -0,0 +1,128 @@
...
+            Bucket.delete(record_bucket.id)
```
nharraud: This one too.

b2share/modules/deposit/api.py
dinosk: to be tested.

b2share/modules/deposit/views.py
```diff
@@ -136,8 +140,14 @@ def records_rest_url_rules(endpoint, list_route=None, item_route=None,
         links_factory=links_factory,
         default_media_type=default_media_type)
 
+    from b2share.modules.records.views import RecordsVersionsResource
+    versions_view = RecordsVersionsResource.as_view(
+        RecordsVersionsResource.view_name.format(endpoint),
+        resolver=resolver)
```
nharraud: This will be removed. We only keep `/api/records/<ID>/versions`. We don't want a duplicate `/api/records/<ID>/draft/versions`

b2share/modules/records/api.py
```diff
@@ -0,0 +1,94 @@
...
+from b2share.modules.deposit.fetchers import b2share_deposit_uuid_fetcher
+from b2share.modules.records.fetchers import b2share_record_uuid_fetcher
+from invenio_indexer.api import RecordIndexer
+from invenio_pidrelations.contrib.versioning import PIDVersioning
+
+class B2ShareRecord(Record):
+    """B2Share record class."""
+    @property
+    def pid(self):
+        """Return an instance of deposit PID."""
```
nharraud: s/deposit/record/

b2share/modules/records/api.py
```diff
@@ -0,0 +1,94 @@
...
+                object_type='rec',
+                # Retrieve the deposit with the Record class on purpose as the current
+                # Deposit api prevents the deletion of published deposits.
+                getter=B2ShareDeposit.get_record,
+            ).resolve(depid.pid_value)
+            deposit.delete()
+
+        # Remove record from versioning chain
+        # if the parent has no other children and no draft child
+        # mark it as deleted
```
nharraud: To test:
- delete last published version, draft exists
- delete last published version, no draft

tests/b2share_unit_tests/records/test_record_versions_rest.py
```diff
@@ -0,0 +1,155 @@
...
+            chain.append(make_record(app, login, data[0]))
+            chain.append(make_record(app, login, data[1], version_of=chain[0]))
+            chain.append(make_record(app, login, data[2], version_of=chain[1]))
+            chain.append(make_record(app, login, data[0], version_of=chain[2]))
+            chain.append(make_record(app, login, data[1], version_of=chain[3]))
+            chain.append(make_record(app, login, data[2], version_of=chain[4]))
+            chain_list.append(chain)
+            print("chain completed")
+
+        time.sleep(1) # wait a bit for elasticsearch
```
dinosk: If we're waiting for the indexing to finish, we could do `current_search_client.indices.flush(*)`

tests/b2share_unit_tests/upgrade/helpers.py
```diff
@@ -36,6 +50,9 @@ def upgrade_run(app):
     runner = CliRunner()
     script_info = ScriptInfo(create_app=lambda info: app)
     return runner.invoke(run, ['-v'], obj=script_info)
+    # from b2share.modules.upgrade.api import upgrade_to_last_version
```
dinosk: to be cleaned

webui/src/components/editrecord.jsx
```diff
@@ -567,6 +567,9 @@ const EditRecord = React.createClass({
         const afterPatch = (record) => {
             if (this.props.isDraft && !this.isForPublication()) {
                 this.props.refreshCache();
+                // TODO(edima): when a draft is publised, clean the state of
```
dinosk: The suggestion in the note imo is better than what we have now, where we clear the cache completely before each fetch.

tests/b2share_unit_tests/records/test_record_versions_rest.py
nharraud: 👍

tests/b2share_unit_tests/upgrade/helpers.py
nharraud: 👍

b2share/modules/records/views.py
```diff
@@ -305,21 +357,77 @@ def put(*args, **kwargs):
...
-        return 'Record was deleted.', 200
+    def __init__(self, resolver=None, **kwargs):
+        """Constructor.
+
+        :param resolver: Persistent identifier resolver instance.
+        """
+        default_media_type = 'application/json'
+        super(RecordsVersionsResource, self).__init__(
+            serializers={
+                'application/json': lambda response: jsonify(response)
```
dinosk: We should use the same serializer as the in the B2ShareRecordResource imo.

b2share/modules/records/views.py
```diff
@@ -305,21 +357,77 @@ def put(*args, **kwargs):
...
+            version_parent_pid_value = pid_versioning.parent.pid_value
+        else:
+            version_parent_pid_value = pid_value
+        records = []
+        pids_and_meta = \
+            db.session.query(PersistentIdentifier, RecordMetadata).join(
+                PIDRelation,
+                PIDRelation.child_id == PersistentIdentifier.id).\
+            filter(PIDRelation.parent_id ==
+                   RecordUUIDProvider.get(version_parent_pid_value).pid.id,
```
nharraud: This should be optimized by making an additional join to the PIDStore table. The current code is doing two requests to the DB instead of one: `RecordUUIDProvider.get(version_parent_pid_value)` + `JOIN query`

PR #1529, opened 2017-08-15
api: record deletion

dinosk:
* ADD record deletion through REST API


emanueldima:
I think we should create a 'delete_record' function in the deposit/api.py, to be used when it's needed, which would be called by the views.py:delete function. Not because I can think of a need to call the delete function in other place, but because this should be the functionality layering: api/http/UI.

tests/b2share_unit_tests/records/test_records_rest.py
```diff
@@ -388,3 +388,75 @@ def test_modify(status, user=None):
...
+                    headers=headers)
+
+                assert request_res.status_code == status
+
+        test_delete(403, creator)
+        test_delete(403, non_creator)
+        test_delete(200, admin)
+
+        with app.test_client() as client:
+            login_user(creator, client)
```
nharraud: This part could have been factorized as it repeats 3 times nearly the same code.

b2share/modules/records/permissions.py
```diff
@@ -53,8 +54,14 @@ def update_record_metadata_need_factory(community):
 update_record_metadata_need = update_record_metadata_need_factory(None)
 
 
+def delete_record_permission_factory(record):
```
nharraud: I think the constructor can be used directly instead.

nharraud:
@emanueldima I agree, it would enable us to add later a CLI function if we want or even do it programmatically in tests.

PR #1530, opened 2017-08-18
global: addition of statistics module

dinosk:
* ADD entrypoint for statistics module, tracking the
      number of file downloads.
    
* ADD celery configuration variables.
   
Signed-off-by: Dinos Kousidis konstantinos.kousidis@cern.ch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,6 @@ matrix:
     - env: REQUIREMENTS=devel REXTRAS=all ES_VERSION=2.3.3
 
 python:
-  - "3.3"
   - "3.4"
   - "3.5"
 
--- a/webui/src/components/editfiles.jsx
+++ b/webui/src/components/editfiles.jsx
@@ -2,6 +2,7 @@ import React from 'react/lib/ReactWithAddons';
 import { Link } from 'react-router'
 import { fromJS, OrderedMap, Map } from 'immutable';
 import moment from 'moment';
+import _ from 'lodash';
 import { serverCache, Error } from '../data/server';
 import { pairs, humanSize } from '../data/misc';
 import { Wait, Err } from './waiting.jsx';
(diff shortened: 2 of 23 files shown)


dinosk:
@emanueldima Hi Emanuel, I am currently working on making the tests pass but I thought to ask you for a review on the UI additions.

webui/src/components/editfiles.jsx
```diff
@@ -518,15 +516,31 @@ export const FileRecordRow = React.createClass({
         return {
             open: false,
             remove: false,
+            downloadsTot: 0,
         };
     },
 
+    componentDidMount() {
```
dinosk: I added the ajax request in componentDidMount here, and called then to update the downloadsTot in the state. If we don't use promises, should I change this to only a callback function?

.travis.yml
```diff
@@ -46,7 +46,6 @@ matrix:
     - env: REQUIREMENTS=devel REXTRAS=all ES_VERSION=2.3.3
 
 python:
-  - "3.3"
```
dinosk: I removed 3.3 because [weakref](https://docs.python.org/3.3/library/weakref.html) was failing trying to call `weakref.finalize`, which I saw doesn't exist in that version.

webui/src/components/editfiles.jsx
emanueldima: Yes, I'd replace it with:
````
serverCache.getFileStatistics(file.version_id, file.bucket, (mystat) => {
    let downloadsTot = mystat.buckets[0].value;
    this.setState({ downloadsTot });
})
````

And in serverCache:
````
getFileStatistics(file_id, bucket_id, successFn) {
    var data = {
        "mystat": {
            "stat": "bucket-file-download-total",
            "params": {
                "start_date": "2010-01-01",
                "bucket_id": bucket_id,
             }
         }
     };
     ajaxPost({
                 url: apiUrls.statistics(),
                 params: data,
                 successFn: response => { 
                     successFn(response.mystat);
                 },
             });
         });
     }
  };		  
};
````

dinosk: @emanueldima changed it to this, it works, thank you

.travis.yml
nharraud: Ok, Invenio is not tested with python 3.3 anymore so we can say that it is not supported.

PR #1535, opened 2017-09-06
Documentation update

thijscobben:
Replaced all doc here by references to EUDAT official documentation site:

https://github.com/EUDAT-Training/B2SHARE-Training


nharraud:
I wouldn't remove the whole doc but rather transform it as "B2SHARE developer" documentation. The training doc will never be able to cover all the information needed by a new developer.

One example would be the `INSTALL.rst`. The first part redirecting to the training doc is fine but it should also explain how to install B2SHARE on bare metal.

emanueldima:
👍

emanueldima:
I found out that not all information from docs/administration.rst is present in the training repository, for instance the small section about updates at the end. We cannot merge this until the training is updated, maybe we can make the PR ourselves. 😞

PR #1537, opened 2017-09-08
ui: fix embargoed record stats

dinosk:
* FIX error in stats query occurring in a record page which has
  an embargo date.


emanueldima:
👍

PR #1548, opened 2017-09-27
Remove direct requirement of celery, update celery config

emanueldima:
Celery is still required by some invenio package.

Fixes:
````
UserWarning: Celery v4 installed, but detected Celery v3 configuration CELERYBEAT_SCHEDULE (use CELERY_BEAT_SCHEDULE instead).
````

### DIFF
--- a/b2share/config.py
+++ b/b2share/config.py
@@ -258,7 +258,7 @@ CELERY_RESULT_BACKEND = "redis://localhost:6379/1"
 #: Accepted content types for Celery.
 CELERY_ACCEPT_CONTENT = ['json', 'msgpack', 'yaml']
 #: Beat schedule
-CELERYBEAT_SCHEDULE = {
+CELERY_BEAT_SCHEDULE = {
     'embargo-updater': {
         'task': 'b2share.modules.records.tasks.update_expired_embargoes',
         'schedule': crontab(minute=2, hour=0),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: b2share/config.py [hunk 1]
```diff
@@ -258,7 +258,7 @@
 #: Accepted content types for Celery.
 CELERY_ACCEPT_CONTENT = ['json', 'msgpack', 'yaml']
 #: Beat schedule
-CELERYBEAT_SCHEDULE = {
+CELERY_BEAT_SCHEDULE = {
```
nharraud: You are sure about this change? It seams that the variable was [renamed in celery 4](http://docs.celeryproject.org/en/latest/userguide/configuration.html#new-lowercase-settings) but not like this, and the previous name is still supported.

nharraud:
Otherwise the `setup.py` changes look good.

b2share/config.py
emanueldima: This actually seems to come from Invenio, see: 
http://pythonhosted.org/Flask-CeleryExt/#celery-v3-and-v4-configuration
https://github.com/inveniosoftware/flask-celeryext/blob/master/flask_celeryext/_mapping.py#L17
https://github.com/inveniosoftware/flask-celeryext/blob/master/flask_celeryext/app.py#L27

nharraud: Ok thanks, I missed that.

PR #1559, opened 2017-10-10
WIP files: B2Safe PID integration

dinosk:
- [x] create a record with b2safe file pids from a data field
- [x] modify the b2safe pids of an existing record with PATCH
- [x] tag the files from b2safe in a record page
- [x] accessing a b2safe pid file redirects to its pid
- [x]  REST API get deposit lists the b2safe pids
- [x] change json structure for the requests like it is indexed in Elasticsearch
- [ ] add b2safe pids to a draft from the UI  (paused)
   - [ ] remove from the UI the button for now
- [ ] query Handle for each PID to verify that it a PID is valid and fetch the checksum (paused)
- [ ] query Handle for each PID to verify that the PID is immutable.
- [ ] update REST API documentation to include this


dinosk:
From the UI the files would be tagged with this badge
screen shot 2017-10-16 at 14 08 09

dinosk:
The UI button in the deposit page could look like this:
screen shot 2017-10-18 at 18 08 53

Popping up a window like this:
untitled drawing

webui/src/components/editfiles.jsx
```diff
@@ -555,7 +558,13 @@ export const FileRecordRow = React.createClass({
                         { this.props.hideFileDownloads ? null :
                             <span className="fileDownloadBadge" style={{marginLeft:'1em', fontSize:11, color: '#888'}}>
                                 <span className="badge" style={{marginLeft:'0.5em'}}> Total Downloads {downloads} </span>
-                            </span> 
+                            </span>
+                        }
+                        {
+                            b2safe_pid ?
```
emanueldima: Why not make the condition just `file.b2safe_pid ?`, without defining the `b2safe_pid`.
Also, probably `file.b2safe_pid` should be boolean, not string.

dinosk: Yes I did both these changes

dinosk:
*Note (to self)*: commit `51bc82d` will be removed and included in a different PR.

webui/src/data/server.js
```diff
@@ -42,6 +42,7 @@ const apiUrls = {
     disciplines()                     { return `${urlRoot}/suggest/disciplines.json` },
 
     statistics()                      { return `${urlRoot}/api/stats` },
+    b2handle_pid_info(file_pid)       { return `${urlRoot}/api/handle/${file_pid}` },
```
nharraud: This should have been rollbacked for now.

webui/src/data/server.js
```diff
@@ -886,7 +889,47 @@ class ServerCache {
...
+        ajaxPost({
+            url: apiUrls.statistics(),
+            params: data,
+            successFn: response => {
+                successFn(response.fileDownloads);
+            },
+        });
+    }
+
+    getB2HandlePidInfo(file_pid, successFn){
```
nharraud: This should have been rollbacked for now.

webui/src/components/editfiles.jsx
```diff
@@ -522,7 +702,7 @@ export const FileRecordRow = React.createClass({
     render() {
         let file = this.props.file;
         file = file.toJS ? file.toJS() : file;
-
+        // const downloads = this.state.downloads[file.key || file.name] || 0;
```
nharraud: This should IMHO be removed.

webui/src/components/editfiles.jsx
```diff
@@ -390,6 +564,12 @@ export const EditFiles = React.createClass({
                             <h3>Add B2DROP files</h3>
                         </button>
                     </div>
+                    <div className="col-md-offset-3 col-md-9" style={{marginBottom:'1em'}}>
+                        <button className='b2dropbutton' onClick={e => this.props.setModal(b2safeZone)}>
```
nharraud: minor issue: We should use a different CSS class for this button.

webui/src/components/editfiles.jsx
```diff
@@ -358,6 +513,24 @@ export const EditFiles = React.createClass({
         if (!this.props.files.length) {
             return false;
         }
+        let record = this.props.record;
+
+        var metadata = this.props.record.get('metadata').toJS();
+        var external_pids = metadata.external_pids;
+        for(var i=0; i < this.props.files.length; i++){
+            var _file = this.props.files[i];
+            console.log('_file:', _file);
```
nharraud: We should remove the logging.

webui/src/components/editfiles.jsx
```diff
@@ -358,6 +513,24 @@ export const EditFiles = React.createClass({
...
+            if(Boolean(external_pids)){
+                for(var j=0; j < external_pids.length; j++){
+                    var external_pid = external_pids[j];
+                    console.log('file key match: ', _file.key);
```
nharraud: We should remove the logging.

webui/src/components/editfiles.jsx
```diff
@@ -358,6 +513,24 @@ export const EditFiles = React.createClass({
...
+                    console.log('external_pid key match: ', external_pid.key);
+                    if(_file.key == external_pid.key){
+                        _file.b2safe = true;
```
nharraud: Isn't the backend already doing this?

webui/src/components/editfiles.jsx
```diff
@@ -390,6 +564,12 @@ export const EditFiles = React.createClass({
                             <h3>Add B2DROP files</h3>
                         </button>
                     </div>
+                    <div className="col-md-offset-3 col-md-9" style={{marginBottom:'1em'}}>
```
nharraud: I thought we planned to disable the UI widget for B2Safe PID addition?

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM dinossimpson/b2safe-integration
```
nharraud: This file should be removed.

b2share/modules/deposit/api.py
```diff
@@ -252,6 +265,35 @@ def commit(self):
...
+            bucket = Bucket.query.filter_by(id=record_bucket.bucket_id).first()
+            object_versions = ObjectVersion.query.filter_by(
+                bucket_id=bucket.id).all()
+            # for the existing files
+            for object_version in object_versions:
+                if object_version.file is None or \
+                        object_version.file.storage_class != 'B':
+                    continue
+                # check that they are still in the file pids list or remove
+                if object_version.key not in self['external_pids']:
```
nharraud: It seems to me that this will always be `true` as `self['external_pids']` is an array of objects.

b2share/modules/records/indexer.py
```diff
@@ -48,6 +48,10 @@ def record_to_index(record):
 def indexer_receiver(sender, json=None, record=None, index=None,
                      **dummy_kwargs):
     """Connect to before_record_index signal to transform record for ES."""
+    if is_deposit(record.model):
+        if 'external_pids' in json['_deposit']:
+            json['external_pids'] = json['_deposit']['external_pids']
```
nharraud: We should be careful to not expose the files "key" if the record is under embargo. This might contain private information. This is why we already remove the `_files` field.

nharraud:
I listed one additional step which is necessary in the final implementation:
=> query Handle server for each PID to verify that the PID is immutable.

b2share/modules/deposit/api.py
```diff
@@ -252,6 +265,35 @@ def commit(self):
...
+                    ObjectVersion.delete(bucket,
+                                         object_version.key)
+                # check that the uri is still the same or update it
+                elif object_version.file.uri != \
+                        self['external_pids'][object_version.key]:
```
nharraud: self[`external_pids`] is not a dict either.

nharraud:
NOTE: because we don't actually created an "external_pids" field in the metadata the PATCH requests are restricted and strange. The "replace" operation will fail and it is impossible to patch the content of the list. It has to be fully replaced with an "add" operation.

b2share/modules/deposit/api.py
```diff
@@ -252,6 +265,35 @@ def commit(self):
...
+                    db.session.query(FileInstance).\
+                        filter_by(key=object_version.key).\
+                        update({"uri": self['external_pids'][object_version.key]})
+            create_b2safe_file(self['external_pids'], bucket)
+            self['_deposit']['external_pids'] = self['external_pids']
```
nharraud: **This is dangerous. This stores anything the user sends without any structure constraint. It will eventually break the indexing when the field types conflict.**

tests/b2share_unit_tests/files/test_b2share_storage.py
```diff
@@ -99,34 +100,35 @@ def test_modify_external_files(app, deposit_with_external_pids,
...
-                "key": "file2.txt",
-                "ePIC_PID": "http://hdl.handle.net/11304/50fafc50-4227-4464-bacc-2d85295c18a7"
-            }, {
-                "key": "file3.txt",
-                "ePIC_PID": "http://hdl.handle.net/11304/50fafc50-4227-4464-bacc-2d85295c18a6"
-            }, {
-                "key": "file4.txt",
-                "ePIC_PID": "http://hdl.handle.net/11304/50fafc50-4227-4464-bacc-2d85295c18a8"
-            }
-        ]
```
nharraud: We should add a comment like: "Check that duplicate files are not allowed"

dinosk:
TODO: add a test for searching (open & closed) records containing external_pids.
(related to https://github.com/EUDAT-B2SHARE/b2share/issues/1636)

nharraud:
Note: Some changes will be done after the Porto meeting for the final release.

Issue #1519, opened 2017-07-19
Improve design consistency of configuration files


thijscobben:
Please not that configuration of B2NOTE takes place in 
https://github.com/EUDAT-B2SHARE/b2share/blob/master/webui/src/components/record.jsx#L13
and in
https://github.com/EUDAT-B2SHARE/b2share/blob/master/webui/src/data/server.js#L277

this should be moved to a config file e.g. https://github.com/EUDAT-B2SHARE/b2share/blob/master/b2share/config.py

Second above config.py file contains "configurable functionality" but also some "internal settings" that should not be exposed to a "technical administrator" who is configuring a b2share instance. Compare B2ACCESS_APP_CREDENTIALS (configurable, but via docker super-environment variables) with DEPOSIT_ENDPOINTS which should not be configurable with CFG_EPIC_USERNAME which should be configurable.

Maybe we should move settings to a file outside docker (for instance postgresql credentials are used in two dockers so should be configured outside/above the Docker containers but B2ACCESS credentials are only used by b2share container but are also configured outside the docker whereas EPIC_USERNAME is also only used by b2share container but configured within the Docker container config.py file?

nharraud:
The `config.py` file is internal and is not supposed to be read by the users. Some time ago [another configuration file](https://github.com/EUDAT-B2SHARE/b2share/blob/d504d6348cdae7868c03a669580ae5d4f7d79810/demo/b2share_demo/config.py) is added with some comments for the user. This of course needs to be improved but it should contain only variables which can be changed by the user.

thijscobben:
This apparantly is implemented and approved and not relevant anymore.

PR #1571, opened 2017-11-06
Small important issues

emanueldima:
- configurable b2note url
- ease the customisation of the UI: moved specific eudat elements in index.html (which can be modified without the need of recompilation) and added a configurable terms-of-use link


emanueldima:
related to https://github.com/EUDAT-B2SHARE/b2share/issues/1519

PR #1584, opened 2017-11-23
fix migrating records issues (overindexing)

emanueldima:


b2share/modules/upgrade/upgrades/upgrade_2_0_0_to_2_1_0.py
```diff
@@ -102,12 +103,20 @@ def alembic_upgrade_database_data(alembic, verbose):
...
+    RecordIndexer.index = lambda s, record: None
+
+    click.secho('migrating deposits.')
     with db.session.begin_nested():
         deposit_pids = PersistentIdentifier.query.filter(
             PersistentIdentifier.pid_type == DepositUUIDProvider.pid_type,
             PersistentIdentifier.status == PIDStatus.REGISTERED,
         ).all()
         for dep_pid in deposit_pids:
+            click.secho('    deposit {}'.format(dep_pid.pid_value))
```
nharraud: Can you remove the pretty printing?

b2share/modules/upgrade/upgrades/upgrade_2_0_0_to_2_1_0.py
```diff
@@ -102,12 +103,20 @@ def alembic_upgrade_database_data(alembic, verbose):
     """Migrate the database data from v2.0.0 to 2.1.0."""
     ### Add versioning PIDs ###
     # Reserve the record PID and versioning PID for unpublished deposits
+
+    # Hack: disable record indexing during record migration
+    from invenio_indexer.api import RecordIndexer
+    old_index_fn = RecordIndexer.index
+    RecordIndexer.index = lambda s, record: None
+
+    click.secho('migrating deposits.')
```
nharraud: Can you remove the pretty printing? 
```
if verbose:
    click.secho('migrating deposits.')
```

b2share/modules/upgrade/upgrades/upgrade_2_0_0_to_2_1_0.py
```diff
@@ -139,6 +148,8 @@ def alembic_upgrade_database_data(alembic, verbose):
                     parent_pid
                 )
                 version_master.update_redirect()
+    click.secho('done migrating deposits.')
```
nharraud: Can you remove the pretty printing? 
```
if verbose:
    click.secho('done migrating deposits.')
```

b2share/modules/upgrade/upgrades/upgrade_2_0_0_to_2_1_0.py
nharraud: Also the message is wrong as it is also migrating published records.

b2share/modules/upgrade/upgrades/upgrade_2_0_0_to_2_1_0.py
emanueldima: I could, but operation is only executed once and takes a long time (1sec/deposit), so it's useful for the users to see what's going on. I would leave it like this.

emanueldima: ... I added the `verbose` test, this may be useful.

PR #1586, opened 2017-11-23
More fixes for oai and ui

emanueldima:


b2share/modules/upgrade/upgrades/upgrade_2_0_0_to_2_1_0.py
```diff
@@ -157,6 +159,14 @@ def alembic_upgrade_database_data(alembic, verbose):
 
 def migrate_record_metadata(record, parent_pid):
     """Migrate a record's metadata to 2.1.0 format."""
+    # Fix bad OAI-PMH identifiers
+    oai_prefix = current_app.config.get('OAISERVER_ID_PREFIX')
+    if oai_prefix:
+        oai_id = record.get('_oai', {}).get('id')
+        if oai_id and not oai_id.startswith(oai_prefix):
+            assert oai_id == record['_deposit']['id']
+            record['_oai']['id'] = oai_prefix + oai_id
```
nharraud: The following code only fixes the "minted" OAI ID which is in the record. Are we sure that the PersistentIdentifier is correct?
You can see the [following line](https://github.com/EUDAT-B2SHARE/b2share/blob/bee5bbab68e8f9fffa47988355a7088e04585e00/b2share/modules/records/minters.py#L73) in our minter. I suspect that if we minted the wrong PIDs, they are also wring in the PersistentIdentifier table.

b2share/modules/records/minters.py
```diff
@@ -77,7 +77,7 @@ def b2share_oaiid_minter(rec_pid, data):
     )
     data.setdefault('_oai', {})
     data['_oai'].update({
-        'id': rec_pid.pid_value,
+        'id': oai_pid_value,
```
nharraud: Why are we not using `[oaiid_minter](https://github.com/inveniosoftware/invenio-oaiserver/blob/master/invenio_oaiserver/minters.py#L38)`? The only thing we do in addition here is set the `sets`. We could just call `oaiid_minter` and then update the sets (even if it is not the best place to do it).
You can see how Zenodo uses it [here](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/minters.py#L79)  and [here](https://github.com/zenodo/zenodo/blob/d0bfb7edd06141483f1c01bc9048a7810cc7ec60/zenodo/config.py#L880)

tests/b2share_unit_tests/upgrade/helpers.py
```diff
@@ -95,6 +95,9 @@ def check_records_migration():
             'type': RecordUUIDProvider.parent_pid_type,
             'value': parent_pid.pid_value,
         })
+        # The OAI-PMH identifier has been modified by the migration
+        if db_record.get('_oai'):
+            exp_record['metadata']['_oai']['id'] = db_record['_oai']['id']
```
nharraud: The OAI ID is deterministic. It would be better to build the expected OAI ID instead of using the one from the DB, otherwise we don't know if we added a bug. We could use a minter on an empty dict and extract the expected PID.

b2share/modules/upgrade/upgrades/upgrade_2_0_0_to_2_1_0.py
emanueldima: The PIDs are correct, you can see here that we're using the right value:
https://github.com/EUDAT-B2SHARE/b2share/blob/bee5bbab68e8f9fffa47988355a7088e04585e00/b2share/modules/records/minters.py#L76

b2share/modules/records/minters.py
emanueldima: As you say, it's basically the same code, it's short and it works, so why change it now? We'd also probably need to mess with some configs.

tests/b2share_unit_tests/upgrade/helpers.py
emanueldima: You're right.

b2share/modules/records/minters.py
nharraud: Yes we can keep it for now. I was just wondering if there was a specific reason as I couldn't find any customization. We can keep this as an issue for the final code cleaning.

emanueldima:
Added some more fixes to make it work with trng and prod data.

PR #1588, opened 2017-11-30
Support new handle server api with b2handle

emanueldima:
Tested with a private test account from SurfSara, check: 
https://epic3.storage.surfsara.nl:8001 with handle 846/f24381a0-821a-4d40-a543-82e76b00f62f

### DIFF
--- a/b2share/modules/deposit/api.py
+++ b/b2share/modules/deposit/api.py
@@ -361,7 +361,8 @@ def create_file_pids(record_metadata):
                            bucket_id=f.get('bucket'), key=f.get('key'),
                            _external=True)
         try:
-            file_pid = createHandle(file_url, checksum=f.get('checksum'))
+            file_pid = createHandle(file_url, checksum=f.get('checksum'),
+                                              fixed=True)
             if file_pid is None:
                 raise EpicPIDError("EPIC PID allocation for file failed")
             f['ePIC_PID'] = file_pid
--- a/b2share/modules/deposit/api.py
+++ b/b2share/modules/deposit/api.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of EUDAT B2Share.
-# Copyright (C) 2016 CERN.
+# Copyright (C) 2016, 2017 University of Tübingen, CERN.
 #
 # B2Share is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -61,6 +61,8 @@ from b2share.modules.schemas.api import CommunitySchema
 from b2share.modules.deposit.minters import b2share_deposit_uuid_minter
 from b2share.modules.deposit.fetchers import b2share_deposit_uuid_fetcher
 from b2share.modules.deposit.providers import DepositUUIDProvider
+from b2share.modules.handle.proxies import current_handle
+from b2share.modules.handle.errors import EpicPIDError
 
 
 class PublicationStates(Enum):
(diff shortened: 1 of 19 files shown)


nharraud:
We will need a migration for the old PIDs.

emanueldima:
@nharraud A B2SHARE migration? What should this do?

nharraud:
If I remember correctly there are now some mandatory fields which we didn't provide in the past. Shouldn't we create them? I don't mean that the migration would change the B2SHARE database. I an talking about a Handle metadata migration.

b2share/modules/records/b2share_epic.py
```diff
@@ -2,18 +2,123 @@
...
 
 from .errors import EpicPIDError
 
 
-def createHandle(location, checksum=None, suffix=''):
+handle_client = None
+handle_prefix = None
+
+
+def init_handle_client(app):
```
nharraud: Why not do the initialization in an extension as we do for other modules?
We could have `b2share.module.handle` with a `proxy.py` providing `current_handle`.

emanueldima: Can you make this change? I don't know what are the advantages of the Invenio way, and just used known Python features.

b2share/modules/records/minters.py
```diff
@@ -93,7 +93,7 @@ def b2share_pid_minter(rec_pid, data):
     throw_on_failure = current_app.config.get('CFG_FAIL_ON_MISSING_PID', True)
 
     try:
-        pid = createHandle(url)
+        pid = create_handle(url)
```
nharraud: At some point we should register the HANDLE PIDs after the commit in an asynchronous task. Otherwise we will have from time to time PIDs which are registered but the transaction rollbacks because of some error.

emanueldima: Agree. Make an issue?

b2share/modules/deposit/api.py
```diff
@@ -361,7 +361,8 @@ def create_file_pids(record_metadata):
                            bucket_id=f.get('bucket'), key=f.get('key'),
                            _external=True)
         try:
-            file_pid = createHandle(file_url, checksum=f.get('checksum'))
+            file_pid = create_handle(file_url, checksum=f.get('checksum'),
```
nharraud: Same as records. At some point we should register the HANDLE PIDs after the commit in an asynchronous task. Otherwise we will have from time to time PIDs which are registered but the transaction rollbacks because of some error.

b2share/modules/records/b2share_epic.py
```diff
@@ -2,18 +2,123 @@
...
+    current_app.logger.info("Created Handle System PID: {}".format(handle))
+
+    CFG_HANDLE_SYSTEM_BASEURL = current_app.config.get(
+        'CFG_HANDLE_SYSTEM_BASEURL')
+    return urljoin(CFG_HANDLE_SYSTEM_BASEURL, handle)
+
+
+def check_eudat_entries_to_handle_pid(handle, fixed=False,
+        checksum=None, checksum_timestamp_iso=None, update=False):
+    """Checks the mandatory EUDAT entries to the Handle PID"""
```
nharraud: It seems to me that this function "migrates" rather than "checks". "Check" is more a synonym of "testing".

nharraud:
@emanueldima I created the B2ShareHandle module and moved the code there. I didn't push the refactoring too far in order to avoid creating new bugs. Later we can cleanly load all the configuration variables in the extension instead of accessing them directly in the functions.

I added a simple test but it does not cover much code. We still need to do manual testing on the servers.

emanueldima:
I rearranged some code to be more symmetric and tested all code paths, so from my point of view it should be good to merge (after commit 92c67d4)

PR #1602, opened 2017-12-14
fix ui bug when searching records after searching drafts

### LINKED ISSUE #1596: UI: blocked on search draft
**Problem:**
Once we start searching for drafts we can't search for records anymore.

**Scenario:**
- login
- go to your profile
- click on "List of your draft records"
- click on the "EUDAT" logo to go back to the main page. => "publication_state:draft" is still in the search bar.
- empty the search bar and click on search. => the search page has an empty search bar but there is still `drafts=1` in the url and drafts are returned instead of published records.

emanueldima:
closes #1596

### DIFF
--- a/webui/src/components/user.jsx
+++ b/webui/src/components/user.jsx
@@ -141,7 +141,7 @@ export const UserProfile = React.createClass({
                     </div>
                     <div className="row">
                         <h3>Own drafts</h3>
-                            <p><Link to={"/records?drafts=1&q=publication_state%3Adraft"}>
+                            <p><Link to={"/records?drafts=1"}>
                                 List of your draft records
                             </Link></p>
                     </div>
@@ -235,7 +235,7 @@ const Token = React.createClass({
                                     <i className="fa fa-trash-o"></i> Remove
                                 </button>
                             </form>
-                    </span> 
+                    </span>
                 </li>;
     }
 });
--- a/webui/src/data/server.js
+++ b/webui/src/data/server.js
@@ -518,7 +518,12 @@ class ServerCache {
 
     searchRecords({q, community, sort, page, size, drafts}) {
         if (community) {
-            q = (q || "") + ' community:' + community;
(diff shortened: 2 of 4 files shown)


webui/src/components/user.jsx
```diff
@@ -141,7 +141,7 @@ export const UserProfile = React.createClass({
                     </div>
                     <div className="row">
                         <h3>Own drafts</h3>
-                            <p><Link to={"/records?drafts=1&q=publication_state%3Adraft"}>
```
nharraud: This one is in fact useful otherwise you will also have published drafts.
We can also change the rest API to not show published drafts.

emanueldima: I'm adding it back only when it's sent to the server, here: https://github.com/EUDAT-B2SHARE/b2share/pull/1602/files#diff-4202620782d33deec0c6acdb6ad427d6R523

nharraud: Strange I get published drafts too when I test it.

webui/src/data/server.js
```diff
@@ -520,6 +520,11 @@ class ServerCache {
         if (community) {
             q = (q || "") + ' community:' + community;
         }
+        if (drafts) {
+            // TODO: change this once the workflows are working.
+            // Add "submitted" drafts.
+            q = (q ? '(' + q + ') && ' : '') + 'publication_state:draft';
```
nharraud: @emanueldima I changed this. The "&&" operator ensures that "publication_state:draft" filters the query as expected, otherwise it is an "or" and any element in the query will bypass it.

emanueldima: You're right, I see now. The same should happen for 'community', the line above

nharraud:
@emanueldima There is still one strange behavior:
- on the main page search with some keywords.
- go to your profile
- go to the draft search page => empty search box
- click on the EUDAT logo to go back to the main page => search box contains the first search query

Issue #1614, opened 2017-12-18
UI/API: bad handling of deleted record versions


nharraud:
**Problems:**
When a record version is deleted the UI does not update correctly:
- [] If the last version is deleted the previous version still warn that it is not the last version.
- [] If the last version is deleted, it is not possible to create a new version via the UI. Via the API it is possible to use the last non-deleted version.
- [] When an intermediate version is selected in the dropdown it shows a 410 error message (which is normal) but doesn't show the dropdown to access the other versions.

**Solution:**
Modify the `/versions` endpoint in order to return the PIDs statuses. Use the information appropriately in the UI.

PR #1618, opened 2017-12-20
WIP ui: dropdown versions menu on deleted records

dinosk:
* ADD the dropdown versions menu on the deleted record pages
  so that you can navigate.

* FIX show the `Create New Version` button only on the last
  non deleted version.
  (Closes https://github.com/EUDAT-B2SHARE/b2share/issues/1614)

### DIFF
--- a/webui/src/data/server.js
+++ b/webui/src/data/server.js
@@ -369,7 +369,6 @@ class ServerCache {
             if (!versionsLink) {
                 return
             }
-            store.setIn(cachePath, null);
             ajaxGet({
                 url: versionsLink,
                 successFn: (data) => {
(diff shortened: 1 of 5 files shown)


webui/src/data/server.js
```diff
@@ -530,6 +531,15 @@ class ServerCache {
         return file;
     }
 
+    getVersions(recordID, callback){
```
emanueldima: In general we should prefer to not use callbacks, but to place the getter results in the store, which automatically triggers a repaint of the UI.

dinosk: I see, I noticed that we follow that pattern but I didn't know the difference. I will change it and create a versions store. Also, we already store the versions in the recordCache, but I couldn't use the record here because if serverCache returns an error when getting a record, the versions attribute is not accessible.

emanueldima: I think I would change the callback here: https://github.com/EUDAT-B2SHARE/b2share/blob/master/webui/src/data/server.js#L435 to test that the return code is 410 (deleted record) and if so, retrieve the versions and place them right in the 'Error' object of that record. You're right we don't have a direct link to versions anymore, which is a pity.

dinosk: Ah yes much better, I will change it as described

PR #1622, opened 2018-01-09
check that all declared presentation items in schemas exist

### LINKED ISSUE #1619: Community specific metadata description missing after schema update
After updating the community specific metadata schema, the specific part is not shown in the schema description page <base>/communities/ . v.2.1.0 and v2.0.1.

emanueldima:
The items in a community schema are displayed in the order defined by the 'presentation' declaration. But if the 'presentation' declares an item which is not actually available in the schema, the UI crashes. This PR checks that all items are present in the schema.

fixes #1619

### DIFF
--- a/webui/src/components/schema.jsx
+++ b/webui/src/components/schema.jsx
@@ -15,8 +15,10 @@ export function getSchemaOrderedMajorAndMinorFields(schema) {
     const presentation = schema.getIn(['b2share', 'presentation']);
     const properties = schema.get('properties');
 
-    const majorIDs = presentation ? presentation.get('major') : null;
-    const minorIDs = presentation ?  presentation.get('minor') : null;
+    const majorIDs = presentation && presentation.get('major') ?
+        presentation.get('major').filter(id => properties.get(id)) : null;
+    const minorIDs = presentation && presentation.get('minor') ?
+        presentation.get('minor').filter(id => properties.get(id)) : null;
 
     let minors = OrderedMap(minorIDs ? minorIDs.map(id => [id, properties.get(id)]) : []);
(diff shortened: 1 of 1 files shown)


webui/src/components/schema.jsx
```diff
@@ -15,8 +15,10 @@ export function getSchemaOrderedMajorAndMinorFields(schema) {
     const presentation = schema.getIn(['b2share', 'presentation']);
     const properties = schema.get('properties');
 
-    const majorIDs = presentation ? presentation.get('major') : null;
-    const minorIDs = presentation ?  presentation.get('minor') : null;
+    const majorIDs = presentation && presentation.get('major') ?
+        presentation.get('major').filter(id => properties.get(id)) : null;
```
nharraud: If `presentation` or `presentation.get('major')` are not set, `majorIDs` is `null` This will create an issue during the call `majorIDs.map(...)` on the following lines. Same for `minorIDs`

nharraud:
Shouldn't we add this to the schema validation when it is added? This won't fix existing schemas but it will avoid to have inconsistent schemas in the future.

webui/src/components/schema.jsx
emanueldima: You're right, they can become null, but we check majorIDs and minorIDs before usage with the ternary operator. The mapping is performed only when they're not null.

nharraud: Ah yes thanks for pointing that out, I read the code too fast.

emanueldima:
added new issue about validation: https://github.com/EUDAT-B2SHARE/b2share/issues/1623

PR #1637, opened 2018-01-17
release: v2.1.1rc1

nharraud:

### DIFF
--- a/b2share/modules/upgrade/upgrades/upgrade_2_1_0_to_2_1_1.py
+++ b/b2share/modules/upgrade/upgrades/upgrade_2_1_0_to_2_1_1.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EUDAT B2Share.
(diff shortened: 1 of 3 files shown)


emanueldima:
@nharraud it may be better not to do the upgrade step, because that would allow us to downgrade to 2.1.0 if necessary. Have there been any db changes?

nharraud:
@emanueldima Why would we need to downgrade? This is an RC so no production should ever be updated with this version?

emanueldima:
If it's a coupling/check between the b2share version and the db version, then fine. I was thinking to allow the minor updates to use older versions of the db, in general.

nharraud:
The upgrades are not only used for upgrading the db, they also add information in the db like "what upgrade has run". This is useful in case we have a known bug in a specific version and we want to have the history of upgrades.
If we don't save all the running versions in the DB we will have no way of knowing "what version of b2share ran just before this upgrade" and do the corresponding fixes.

Issue #1632, opened 2018-01-12
PID migration failure


emanueldima:
The command `check_and_update_handle_records` used to migrate the PIDs to the new EUDAT format, failed on production with this trace:

````
# b2share b2records check_and_update_handle_records -v
checking PIDs for all records
Traceback (most recent call last):
  File "/usr/bin/b2share", line 11, in <module>
    load_entry_point('b2share', 'console_scripts', 'b2share')()
  File "/usr/lib/python3.5/site-packages/click/core.py", line 716, in call
    return self.main(*args, **kwargs)
  File "/usr/lib64/python3.5/site-packages/flask/cli.py", line 345, in main
    return AppGroup.main(self, args, *kwargs)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.5/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.5/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), args, *kwargs)
  File "/usr/lib64/python3.5/site-packages/flask/cli.py", line 229, in decorator
    return __ctx.invoke(f, args, *kwargs)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/eudat/b2share/b2share/modules/records/cli.py", line 73, in check_and_update_handle_records
    for record in list_db_published_records():
  File "/eudat/b2share/b2share/modules/records/utils.py", line 59, in list_db_published_records
    if is_publication(record.model):
  File "/eudat/b2share/b2share/modules/records/utils.py", line 42, in is_publication
    return record.json['$schema'].endswith('#/json_schema')
TypeError: 'NoneType' object is not subscriptable
````

emanueldima:
solved by https://github.com/EUDAT-B2SHARE/b2share/pull/1649

PR #1649, opened 2018-02-14
ignore deleted records in list_db_published_records

emanueldima:


emanueldima:
should fix https://github.com/EUDAT-B2SHARE/b2share/issues/1632

nharraud:
If this was the problem it was probably because of deleted records as they have have no JSON.

Issue #1647, opened 2018-02-13
search: disable date_detection and numeric_detection from elasticsearch mapping


nharraud:
**Problem:**
The current record and deposit mappings are letting elasticsearch detect date and numeric fields. This makes the indexing fail if some community-specific fields look like date at first but have later an invalid format.
After testing a dump of the database it looks like this is already impacting production.

**Solution:**
Modify the record and deposit mappings [by disabling date_detection and numeric_detection](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dynamic-field-mapping.html)

Add those 2 lines in both `b2share/modules/deposit/mappings/deposits/deposits.json` and `b2share/modules/records/mappings/records/records.json`
```diff
"mappings" : {
      "record" : {
+          "date_detection": false,
+          "numeric_detection": false,
```

```diff
"mappings" : {
      "deposit" : {
+          "date_detection": false,
+          "numeric_detection": false,
```

Test it by running:
`b2share index destroy`
`b2share index init`
`b2share index reindex`
`b2share index run`

No exception should be returned. This should be tested with the latest prod database dump as there is one record which raises an exception when it is indexed.

Check that the statistic indices are not destroyed during the reindexing. This has already been tested before but it's better to do it again just in case.

**Priority:**
**This needs to be fixed before the 2.1.1 release** or b2share.eudat.eu will have difficulties when indexing records.

emanueldima:
solved by https://github.com/EUDAT-B2SHARE/b2share/pull/1650

PR #1650, opened 2018-02-14
disable automatic elasticsearch date and numeric detection

emanueldima:
Tested manually, it now works on prod db.

solves https://github.com/EUDAT-B2SHARE/b2share/issues/1647


nharraud:
I think this PR fails because [`embargo_date`](https://github.com/EUDAT-B2SHARE/b2share/issues/1647) is not defined as a date.

emanueldima:
Added the `embargo_date` mappings to pass `test_update_expired_embargo`

PR #1657, opened 2018-02-21
add cli command to add missing handles to a record and its files

emanueldima:


b2share/modules/records/cli.py
```diff
@@ -102,6 +104,49 @@ def check_and_update_handle_records(update, verbose):
                         click.secho('  file PID ok: {}'.format(pid))
 
 
+@b2records.command()
+@with_appcontext
+@click.option('-u', '--update', is_flag=True, default=False)
+@click.argument('record_id', required=True)
+def check_handles(update, record_id):
+    """Allocates handles for a record and its files, if necessary"""
+    rec_pid = RecordUUIDProvider.get(pid_value=record_id).pid
```
nharraud: `record_id` confuses me here. The record id is normally its internal uuid, but here it seems to be the value of the b2rec PID pointing to the record. Those are two different things. If it is the record's b2rec pid value we could name this variable `record_pid` or `b2rec_pid`.

b2share/modules/records/cli.py
```diff
@@ -102,6 +104,49 @@ def check_and_update_handle_records(update, verbose):
                         click.secho('  file PID ok: {}'.format(pid))
 
 
+@b2records.command()
+@with_appcontext
+@click.option('-u', '--update', is_flag=True, default=False)
+@click.argument('record_id', required=True)
+def check_handles(update, record_id):
+    """Allocates handles for a record and its files, if necessary"""
```
nharraud: s/Allocates/Allocate/
add `.` at the end.

b2share/modules/records/cli.py
```diff
@@ -102,6 +104,49 @@ def check_and_update_handle_records(update, verbose):
...
+        click.secho('record {} has no handle'.format(record_id), fg='red')
+        if update:
+            b2share_pid_minter(rec_pid, record)
+            record_updated = True
+            click.secho('    handle added to record', fg='green')
+        else:
+            click.secho('use -u argument to add a handle to the record')
+
+    files_ok = True
+    for f in record.get('_files', []):
```
nharraud: We should check what happens here for external files referenced by PID.

b2share/modules/records/cli.py
```diff
@@ -102,6 +104,49 @@ def check_and_update_handle_records(update, verbose):
                         click.secho('  file PID ok: {}'.format(pid))
 
 
+@b2records.command()
+@with_appcontext
+@click.option('-u', '--update', is_flag=True, default=False)
+@click.argument('record_id', required=True)
+def check_handles(update, record_id):
```
nharraud: Do we want to keep `check_and_update_handle_records` above this function? Both functions check the PIDs, seems redundant to me.

emanueldima: `check_and_update_handle_records` adds the EUDAT/xx entries to pid handles, it has a different purpose. `check_handles` adds missing pids to a record and its files.

b2share/modules/records/cli.py
emanueldima: https://github.com/EUDAT-B2SHARE/b2share/issues/1658

emanueldima: https://github.com/EUDAT-B2SHARE/b2share/pull/1659

Issue #1658, opened 2018-02-21
Don't allocate handle PIDs for external files


emanueldima:
In v2.1.1 B2SHARE allocates handles for external files. As the external files should already have been assigned to the record via a PID, this behaviour is incorrect.

emanueldima:
solved by https://github.com/EUDAT-B2SHARE/b2share/pull/1659

PR #1659, opened 2018-02-21
don't allocate handle pids for external files

emanueldima:
solves https://github.com/EUDAT-B2SHARE/b2share/issues/1658

### DIFF
--- a/tests/b2share_unit_tests/records/test_records_api.py
+++ b/tests/b2share_unit_tests/records/test_records_api.py
@@ -202,3 +202,39 @@ def test_record_delete_version(app, test_records, test_users):
 
         # TODO: test in ES that the previous version is reindexed
         # and can be found
+
+
+def test_record_publish_adds_no_handles_for_external_files(app,
+                            records_data_with_external_pids,
+                            test_records_data):
+    """Test that no handle PIDs are created for external files."""
+    for metadata in test_records_data:
+        with app.app_context():
+            app.config.update({'FAKE_EPIC_PID': True})
+
+            external_pids = records_data_with_external_pids['external_pids']
+            external_dict = {x['key']: x['ePIC_PID'] for x in external_pids}
+            data = deepcopy(metadata)
+            data['external_pids'] = deepcopy(external_pids)
+
+            record_uuid = uuid.uuid4()
+            b2share_deposit_uuid_minter(record_uuid, data=data)
+
+            deposit = Deposit.create(data, id_=record_uuid)
+            ObjectVersion.create(deposit.files.bucket, 'real_file_1.txt',
+                             stream=BytesIO(b'mycontent'))
+            ObjectVersion.create(deposit.files.bucket, 'real_file_2.txt',
+                             stream=BytesIO(b'mycontent'))
+            deposit.submit()
+            deposit.publish()
+            deposit.commit()
+
+            _, record = deposit.fetch_published()
+
+            # external files don't get a handle PID, they already have one
+            # which is stored in record['_deposit']['external_pids']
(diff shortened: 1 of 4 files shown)


emanueldima:
@nharraud Tests added

tests/b2share_unit_tests/records/test_records_api.py
```diff
@@ -202,3 +202,40 @@ def test_record_delete_version(app, test_records, test_users):
...
+            deposit.publish()
+            deposit.commit()
+
+            _, record = deposit.fetch_published()
+
+            # external files don't get a handle PID, they already have one
+            # which is stored in record['_deposit']['external_pids']
+            for f in record.files:
+                if f['key'] in external_keys:
+                    assert f.get('ePIC_PID') is None
```
nharraud: Won't the UI be missing the ePIC_PID in `_files`? We could also just copy the PID from external_pids.

emanueldima: That is true. Right now we don't show any pids for external files. The tests would have to be a bit different then.

nharraud: We can just change the PR by copying the PID from external_pids, which simplifies the work of 3rd party scripts. They will always be able to retrieve the ePIC_PID in `_files` that way.

emanueldima:
I changed the serialization function to add the ePIC_PID of external files into the files section. With this change the UI works as expected.

tests/b2share_unit_tests/records/test_records_rest.py
```diff
@@ -520,3 +520,46 @@ def test_access(user=None, deleted=True):
...
+
+            for f in record['files']:
+                assert f['ePIC_PID']
+                if f['key'] in uploaded_files:
+                    # uploaded (internal) file
+                    assert '0000' in f['ePIC_PID'] # freshly allocated fake pid
+                else:
+                    # external file
+                    assert f['b2safe']
+                    assert '0000' not in f['ePIC_PID'] # external pid
```
nharraud: Why not testing directly if the PID is the same?
```
assert f['ePIC_PID'] == [rec for rec in records[external_pids] if rec['key'] == f['key']][0]['ePIC_PID']
```

emanueldima: yep, fixed

tests/b2share_unit_tests/records/test_records_rest.py
```diff
@@ -520,3 +520,48 @@ def test_access(user=None, deleted=True):
...
+                assert f['ePIC_PID']
+                if f['key'] in uploaded_files:
+                    # uploaded (internal) file
+                    assert '0000' in f['ePIC_PID'] # freshly allocated fake pid
+                else:
+                    # external file
+                    assert f['b2safe']
+                    external_pid = [rec for rec in external_pids
+                                    if rec['key'] == f['key']]
+                    assert f['ePIC_PID'] == external_pid[0]['ePIC_PID']
```
nharraud: There are two different external PIDs in this record: https://github.com/EUDAT-B2SHARE/b2share/blob/master/tests/conftest.py#L352-L359

How is it possible that every file in record['files'] has the same external PID value?

emanueldima: `external_pid` is the one that matches the current file (defined just above). It's distinct from `external_pids` (it's not a great name, I admit).

emanueldima: Actually it's better to just change this name.

nharraud:
Thanks for the changes :)

PR #1688, opened 2019-01-30
demo: overlay configuration using invenio_config

### LINKED ISSUE #1686: create a demo record with fake DOI
Currently demo records created with `b2share demo` don't have a fake DOI assigned to them.
It would be useful to have a demo record with DOI, e.g. for testing DataCite serializer.

stefan-wolfsheimer:
* Enable the same configuration overlay mechanism as in b2share main code

* Properties in demo/b2share_demo/config.py can be overwritten with
  env variables with the prefix B2SHARE_DEMO

* example to enable fake PID / DOI:
  export B2SHARE_DEMO_CFG_FAIL_ON_MISSING_PID=1
  export B2SHARE_DEMO_CFG_FAIL_ON_MISSING_DOI=1
  export B2SHARE_DEMO_FAKE_EPIC_PID=1
  export B2SHARE_DEMO_FAKE_DOI=1

closes #1686

### DIFF
--- a/demo/b2share_demo/cli.py
+++ b/demo/b2share_demo/cli.py
@@ -34,9 +34,14 @@ from flask.cli import with_appcontext
 from flask import current_app
 from invenio_db import db
 from invenio_files_rest.models import Location
+from invenio_config import create_conf_loader
 
 from .helpers import load_demo_data
 from . import config as demo_config
+
+
+config_loader = create_conf_loader(config=demo_config,
+                                   env_prefix="B2SHARE_DEMO")
 
 
 @click.group(chain=True)
@@ -50,6 +55,7 @@ def demo():
 def load_data(verbose):
     """Load demonstration data."""
     # add files location
+    config_loader(app=current_app)
     files_path = os.path.join(current_app.instance_path, 'files')
     if os.path.exists(files_path):
         rmtree(files_path)


hjhsalo:
There are two possible ways to do this. I\m not sure if both or just one will work.

1\. Set `FAKE_EPIC_PID` and `FAKE_DOI` to `True` when running `b2share demo` commands
- This might work since [`helpers.py`](https://github.com/EUDAT-B2SHARE/b2share/blob/03d490120bfb096d102cb2c9d9f38bf2b05cc524/demo/b2share_demo/helpers.py#L232) utilizes [`b2share_record_uuid_minter()`](https://github.com/EUDAT-B2SHARE/b2share/blob/03d490120bfb096d102cb2c9d9f38bf2b05cc524/b2share/modules/records/minters.py#L43), and [b2share_pid_minter()](https://github.com/EUDAT-B2SHARE/b2share/blob/03d490120bfb096d102cb2c9d9f38bf2b05cc524/b2share/modules/records/minters.py#L87) takes `FAKE_EPIC_PID` value into account as well (https://github.com/EUDAT-B2SHARE/b2share/blob/03d490120bfb096d102cb2c9d9f38bf2b05cc524/b2share/modules/handle/api.py)

2a\. Add PID and DOI information to demo records:
- https://github.com/EUDAT-B2SHARE/b2share/tree/master/demo/b2share_demo/data/records
- https://github.com/EUDAT-B2SHARE/b2share/blob/master/demo/b2share_demo/data/records/rda/47077E3C-4B9F-4852-A407-09E338AD4620.json

2b\. Implement necessary code changes to `helpers.py`:
- https://github.com/EUDAT-B2SHARE/b2share/blob/03d490120bfb096d102cb2c9d9f38bf2b05cc524/demo/b2share_demo/helpers.py#L232

Could you test the 1st option and in case it doesn't work, I can check the necessary changes to implement option 2_

stefan-wolfsheimer:
B2SHARE_DEMO_FAKE_EPIC_PID=1
B2SHARE_DEMO_FAKE_DOI=1
works without the change in the cli loader. Validation against the datacite schema was positive.
I cannot reproduce why it did not work in the first place. We can disregard this PR.

PR #1784, opened 2020-04-30
Schema unit

### LINKED ISSUE #1782: Allow definition of unit for schema fields
In metadata schemas it should be possible to add the unit of the field. An example field would look like this:

```
        "s_dec": {
            "type": "number",
            "title": "Declination",
            "description": "",
            "unit": "Radians J2000"
        }
```

The unit can be shown on the form and on the landing page of a record.

hevp:
close #1782


hjhsalo:
Two commits with same name in commit history. Maybe squash and/or rename?
Otherwise LGTM.

hjhsalo:
Rebase on top of latest master also

PR #1869, opened 2021-12-01
Fix OAuth2 special char issue and update B2ACCESS test server address

### LINKED ISSUE #1867: Update B2ACCESS configuration
The staging instance of B2ACCESS is not longer available. Please update your code to the integration instance. You just need to replace the domain to b2share-integration.fz-juelich.de. The used endpoints/paths are the same.

### LINKED ISSUE #1868: Bug: Special chars in OAuth2 secret
B2SHARE's OAuth2 client secret currently can't contain %-signs.

This worked up to around 11.12.2019, but failed in production unexpectedly. Most likely changes in B2ACCESS' Unity version has brought the hidden problem up.

Problem seems to be that the body/payload of the request to B2ACCESS is not being URL encoded. This is a requirement per RFC 6749 [section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1). [Appendix B](https://datatracker.ietf.org/doc/html/rfc6749#appendix-B) also shows that %-sign is reserved for escaping purposes. See [Wikipedia](https://en.wikipedia.org/wiki/Percent-encoding) for human readable details.

JohannesLares:
B2SHARE's B2ACCESS OAuth2 implementation don't support %-sign in client secret. The problem is `b2access_basic_auth` variable, where client secret has not been encoded properly. Added `urllib.parse.quote` function to encode client secret.

Added `Content-Type` header for good practice. The header can also be found in [flask-oauthlib/client.py](https://github.com/lepture/flask-oauthlib/blob/master/flask_oauthlib/client.py#L666) line 666. B2SHARE implementation overrides these functionalities and thus the header is not automatically added.

B2ACCESS test server has been changed to integration B2ACCESS server due to removal of old test server. Changed server address to B2SHARE config.py file.

closes #1868 , closes #1867

### DIFF
--- a/b2share/config.py
+++ b/b2share/config.py
@@ -205,7 +205,7 @@ B2ACCESS_APP_CREDENTIALS = dict(
 
 B2ACCESS_BASE_URL = 'https://b2access.eudat.eu/'
 if os.environ.get("USE_STAGING_B2ACCESS"):
-    B2ACCESS_BASE_URL = 'https://unity.eudat-aai.fz-juelich.de/'
+    B2ACCESS_BASE_URL = 'https://b2access-integration.fz-juelich.de'
 
 OAUTHCLIENT_REMOTE_APPS = dict(
     b2access=make_b2access_remote_app(B2ACCESS_BASE_URL)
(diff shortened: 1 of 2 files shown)


hjhsalo:
LGTM. I will merge this.

hjhsalo:
Please rebase on top of latest master

PR #1887, opened 2022-12-22
Fix citation rendering for safari

JohannesLares:
There have been an issue with fetching the citation on Safari browser (Mac, iOS, iPad). Issue originates from not accepted CORS header. Fix by fetching citation directly from data.crosscite.org.


hjhsalo:
LGTM

PR #1888, opened 2022-12-22
Update from B2HANDLE to pyhandle library

hjhsalo:
Change B2HANDLE library to pyhandle library since B2HANDLE has been deprecated.


JohannesLares:
LGTM

PR #1896, opened 2023-03-22
Create Statistics API

gek20:
* API check using AUTH-TOKEN
* return total quota
* return users info

### DIFF
--- a/b2share/modules/stats/views.py
+++ b/b2share/modules/stats/views.py
@@ -31,6 +31,8 @@ from flask import Blueprint, request,  jsonify, make_response
 from invenio_rest import ContentNegotiatedMethodView
 from invenio_oauth2server.decorators import require_api_auth
 from b2share.modules.access.permissions import admin_only
+from invenio_access.permissions import superuser_access
+from b2share.modules.access.permissions import StrictDynamicPermission
 
 from .api import get_quota, users_list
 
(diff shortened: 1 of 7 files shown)


b2share/modules/stats/views.py
```diff
@@ -0,0 +1,84 @@
...
+    }
+    try:
+        return actions[value]()
+    except KeyError:
+        return None
+
+
+
+class B2shareStatisticsCollection(ContentNegotiatedMethodView):
+    """Ownership Management records."""
```
hjhsalo, line 53: Copy-paste?

hjhsalo: Maybe something like this?
```
    """B2SHARE specific stats endpoint.
    
    This is not an extension to invenio-stats, but an independent
    endpoint in B2SHARE.
    """
```

b2share/modules/stats/views.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """Constructor."""
+        super(B2shareStatisticsCollection, self).__init__(
+            default_method_media_type={
+                'GET': 'application/json',
+            },
+            default_media_type='application/json', *args, **kwargs)
+
+    @require_api_auth()
```
hjhsalo, line 65: Change to superuser permissions only?
Maybe this works? `@admin_only.require(http_exception=403)` when you do import with `from b2share.modules.access.permissions import admin_only`

PR #1905, opened 2023-05-03
Update B2SHARE Dockerfile

JohannesLares:
Install specific nodejs version to fix issue, where specific electron version can not be installed.


gek20 (approved):
LGTM

PR #1910, opened 2023-06-13
Fix imports in record management commands

gcakir:
The required function imports were forgotten or excluded. Now, they are added with this commit. @hjhsalo

### DIFF
--- a/b2share/modules/records/cli.py
+++ b/b2share/modules/records/cli.py
@@ -23,7 +23,6 @@
 
 """B2Share cli commands for records."""
 
-
 from __future__ import absolute_import, print_function
 
 import click
@@ -45,11 +44,15 @@ from b2share.modules.records.tasks import update_expired_embargoes \
     as update_expired_embargoes_task
 from .utils import list_db_published_records
 from b2share.modules.handle.proxies import current_handle
+from flask import current_app
+
+from b2share.utils import get_base_url
 
 
 @records.group()
 def manage():
     """B2SHARE record management commands."""
+
 
 @manage.command()
 @with_appcontext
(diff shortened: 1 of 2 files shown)


b2share/modules/records/cli.py
```diff
@@ -45,12 +44,17 @@
     as update_expired_embargoes_task
 from .utils import list_db_published_records
 from b2share.modules.handle.proxies import current_handle
+from flask import current_app
+from flask.cli import with_appcontext
```
JohannesLares, line 48: This seems to be a duplicate of line 29.

JohannesLares (changes requested):
It is true, that imports for `current_app` and `get_base_url` were missing.

Please, remove duplicate import (separate comment)

JohannesLares:
And also, if you could squash all three commits to a single one.

gcakir:
I squashed the commits and removed the duplicates. Could you please review it again?

b2share/modules/records/cli.py
```diff
@@ -45,12 +44,16 @@
     as update_expired_embargoes_task
 from .utils import list_db_published_records
 from b2share.modules.handle.proxies import current_handle
+from flask import current_app
+
+from ...utils import get_base_url
```
JohannesLares, line 49: It would be better to use exact path here and not `...utils` to make it clearer, but this does not affect review.

JohannesLares (approved):
Looks good to me. Can be rebased on top of the master. One change request that could be done, if wanted.

b2share/modules/records/cli.py
gcakir: The exact location is now changed to `b2share.utils` with the latest commit.

gcakir: This should now also be changed to `b2share.utils`.

JohannesLares:
LGTM => Rebase & Merge

PR #1916, opened 2023-08-26
Blur community description when it's too long

gcakir:
It will only execute if the text is over 150 characters long. For example:

Screenshot from 2023-08-26 23-31-48

@hjhsalo @JohannesLares


hjhsalo:
Looks good to me.

PR #1917, opened 2023-08-26
Load fonts locally instead of using Google API

gcakir:
Adding changes that implement the loading of the fonts locally instead of using Google API


hjhsalo:
Looks good to me.

PR #1919, opened 2023-08-28
Fix the parsing of ePIC PIDs

gcakir:
If the variable hdl is, for example, `http://hdl.handle.net/21.T11975/b85684f9-5e24-4731-a6e2-94a8d277b838`, with the original code with `[3:]`, the output would be an empty string. With the updated code with `[-2:]`, the output would be:
`21.T11975/b85684f9-5e24-4731-a6e2-94a8d277b838`

@hjhsalo @JohannesLares


hjhsalo:
Looks good to me.

PR #1920, opened 2023-10-31
Add User-Agent Header to Request Object

gek20:
- add "User-Agent" Header to urllib.Request object


hjhsalo:
Looks good to me. No need define config.py or any examples for DEFAULT_USER_AGENT since I doubt that this would be ever change in B2SHARE v2.

PR #1921, opened 2023-11-01
Fix PATCH record for TypeError when missing doi

gek20:
- fix bug where  missing doi raises TypeError TypeError: 'NoneType' object is not subscriptable'
- log warning if doi is missing


hjhsalo:
Looks good to me.

PR #1928, opened 2024-03-04
Add cli commad list_root_schemas

gek20:
- Check if the root_schemas are loaded in the db
- Check if the root_schemas loaded in the db are also in the folder root_schemas
- Check if the root schemas are valid and up to date


JohannesLares:
Tests missing

PR #1935, opened 2024-07-22
Add multistage build and update requirements

gek20:


dockerize/Dockerfile
```diff
@@ -102,13 +86,25 @@ COPY dockerize/supervisord.conf /etc/
...
 #
 
-WORKDIR /eudat/b2share
-ADD webui webui
-WORKDIR /eudat/b2share/webui
-RUN node_modules/webpack/bin/webpack.js -p
-
 WORKDIR /eudat
 ADD . b2share
+RUN rm -rf /eudat/b2share/webui
```
hjhsalo, line 91: I assume that this done because the COPY command on line 107 fails to write (or overwrite) the /eudat/b2share/webui ?
Maybe add a line comment?
Something like "# Remove webui code brought in with 'ADD . b2share' since webui build is done in its own stage.

dockerize/Dockerfile
```diff
@@ -102,13 +86,25 @@ COPY dockerize/supervisord.conf /etc/
 # Copy the rest of B2Share files
 #
 
-WORKDIR /eudat/b2share
-ADD webui webui
-WORKDIR /eudat/b2share/webui
-RUN node_modules/webpack/bin/webpack.js -p
-
 WORKDIR /eudat
 ADD . b2share
```
hjhsalo, line 90: Is this intentionally here after line 75 & 76?
If I understood the Dockerfile content correctly, line 76 installs B2SHARE as editable from a folder which does not contain b2share source code at the time of install.

hjhsalo:
Commit history needs to be squashed.
Maybe add a reference to CSCEUDAT Jira ticket(s) to the commit message, just in case someone needs to check the comments in there in the future.

dockerize/Dockerfile
RaikasNeiti: Comments have been added to make the stages more clear

dockerize/Dockerfile
RaikasNeiti: Yes it is intentional. It's the same as it was in the orginal b2share dockerfile https://github.com/EUDAT-B2SHARE/b2share/blob/master/dockerize/Dockerfile#L86

gek20:
LGTM now.

PR #1936, opened 2024-07-29
Release version 2.4.0

gek20:
Updates after version 2.3.4:
- New Dockerfile with multiphase build
- New Eudat logos
- Update requirements


hjhsalo:
With `ENV LANG=en_US.UTF-8` I get an error from Python Click, since en_US locale is not installed in python-slim image:
```
$ locale -a
C
C.UTF-8
POSIX
```

So change to `ENV LANG=C.UTF-8` or install en_US locale in Dockerfile.
I am not 100% sure if B2SHARE needs en_US for something else, but I doubt it, so change to `ENV LANG=C.UTF-8`.

PR #1939, opened 2024-09-27
Upgrade psycopg version where found

plaihonen:
At minimum libpq 10 is needed for connecting to the database with scram encryption enabled for logins.


gek20:
LGTM; Creation of record tested.

PR #1942, opened 2024-11-07
Fix user menu bug with safari

JohannesLares:
Safari does not open the menu, when clicking email. This is now fixed.
Added a "feature" (or fixed another bug), where clicking the email when menu visible did not collapse the menu.
Tested on Safari and Chrome.


gek20:
LGTM

gek20 (approved):
Tested.

